PR #6, opened 2015-03-16
Add Qt5 Support

Tobias-Fischer:
Hi,

I want to use rtabmap with Qt5, and this pull request contains the changes which are needed to get it running under Ubuntu. I know that this definitely needs more work to generalize to Windows and Mac, however I am not using either of those platforms.

What I changed is:
1. Update header files (e.g. QtGui/QWidget -> QWidget)
2. Update CMakeFiles - I created an option to specify the version of QT, and updated all the QT related commands and linker stuff
3. Update some functions, as they are deprecated in Qt5.
4. I also updated one function related to PCL 1.8 (with a #if/#elseif directive)

Please note that PCL as well as VTK have to be compiled with Qt5 (see http://www.vtk.org/Wiki/VTK/Configure_and_Build#Qt5..2A)

Some issues which might arise:
1. I am not sure which cmake version is needed for the Qt5 stuff. It could be either 2.8.9 or 2.8.11, but certainly higher than the current minimum version (2.8.5)
2. Hopefully all the header files are working fine with both, Qt4 and Qt5
3. Hopefully the changes of the deprecated functions in Qt5 are compatible with Qt4, but I am not completely sure.

Please let me know whether you find this contribution useful, and what changes you want me to make to merge it to the main repository. For the beginning, it would be good to know whether something is broken when using Qt4 (I cannot/don't want to test it, as I would need to recompile VTK+PCL ..)

Best,
Tobias


matlabbe:
I've merged with a little change for Qt4 on QPrinter include. I noticed that QT5_USE_MODULES is not set in some tools, is it because it was called on the rtabmap_gui target so we don't have to set it again on projects depending on rtabmap_gui? I can't test on Qt5 for now.

Tobias-Fischer:
Hi @matlabbe,
Thanks for the quick merge! That will be useful for me, as I don't have to hold a separate version which I have to keep updated. I'll make sure to fix any issues related to Qt5.

Regarding your questions:
1) Yes, I was surprised myself that I don't have to set QT5_USE_MODULES for the tools. However, it works nicely for me - probably due to the link to rtabmap_gui. If you consider it to be more clean, feel free to add it in the tools/examples though.
2) Actually, the `#if / #elseif` is not needed. `#include <QPrinter>` works nicely due to `QT5_USE_MODULES(rtabmap ... PrintSupport)` in QT5 as well. I did not know about this at first, so I changed it to `#include <QtPrintSupport/QPrinter>` and forgot to change it back. Sorry for that, would you mind fixing it in one of the next commits?

Thanks again!
Tobias

Tobias-Fischer:
Oh, and by the way, I messed up your `.gitignore` in `./build`. Sorry for that! I find it much cleaner to ignore the whole folder rather than individual files in the folder, so I added `./build` in my local git exclude ...

matlabbe:
Fixed in https://github.com/introlab/rtabmap/commit/5ea1a0fe1459ca7ab9e911da04a12a91ab874cd1 .

Tobias-Fischer:
Great, thanks!

PR #108, opened 2016-08-26
Improve Freenect2 performance

lesterlo:
I added 2 modification to improve the Freenect2 performance and fixed the color issue in TegraJPEG.

1. Added Freenect2 cuda pipeline.
2. Fixed TegraJPEG Color problem, convert from RGB to BGR.

I have already tested the code in Nvidia TX1.


matlabbe:
Looks good! thx

PR #110, opened 2016-08-30
in FindDC1394.cmake Mac OS X based framework setup gives errror.

berkaydeniz:
-framework CoreServices is redundant and causes the build the fail.


matlabbe:
On my system (Mac OS X 10.9.5), with or without the pull request it works. I'll merge it, thx!

PR #184, opened 2017-04-18
change for zed sdk 2.0

akirayou:
ZED SDK version is 2.0 now.(https://www.stereolabs.com/developers/)
So I want to use 2.0. 
It seems works well for me (Ubuntu 16.04).

There may be better solution in following 2 implementations. Anyway it works.
-slMat3cvMat  in line 939
-for loop in line 1016


matlabbe:
Thx!

PR #194, opened 2017-05-10
Depth calibration: using appropriate bin_width and bin_height for all cameras

T0ny0:
`DiscreteDepthDistortionModel` constructor has default arguments:
`DiscreteDepthDistortionModel(int width, int height, int bin_width = 8, int bin_height = 6, double bin_depth = 2.0, int smoothing = 1);`
And inside this constructor there is an assertion:
`UASSERT(width_ % bin_width_ == 0);`
`UASSERT(height_ % bin_height_ == 0);`
Assertion fails on Kinect v2 frames [512*424]: `424 % 6 != 0`
So I added method for choosing bin_height and bin_width size.

### DIFF
--- a/corelib/include/rtabmap/core/clams/discrete_depth_distortion_model.h
+++ b/corelib/include/rtabmap/core/clams/discrete_depth_distortion_model.h
@@ -67,6 +67,16 @@ namespace clams
   class RTABMAP_EXP DiscreteDepthDistortionModel
   {
   public:
+    // returns all divisors of num
+    static std::set<size_t> getDivisors(const size_t &num);
+
+    // returns divisor from divisors closest to ref
+    static size_t getClosestToRef(const std::set<size_t> &divisors, const double &ref);
+
+    // sets bin_width and bin_height to appropriate values
+    static void getBinSize(const size_t &width, const size_t &height, size_t &bin_width, size_t &bin_height);
+
+  public:
     DiscreteDepthDistortionModel() :
     	width_(0),
 		height_(0),
--- a/corelib/include/rtabmap/core/util3d.h
+++ b/corelib/include/rtabmap/core/util3d.h
@@ -319,6 +319,15 @@ pcl::PointCloud<pcl::PointXYZ>::Ptr RTABMAP_EXP loadCloud(
 		int downsampleStep = 1,
 		float voxelSize = 0.0f);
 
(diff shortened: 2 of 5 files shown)


matlabbe:
Indeed, there is a problem with kinectv2 depth images. So in your example the `bin_width` will be chosen as 4?
Before merging, can you move the methods in `clams` namespace instead of `util3d` namespace, as it is quite specific to clams. Maybe in [discrete_depth_distortion_model.h](https://github.com/introlab/rtabmap/blob/master/corelib/include/rtabmap/core/clams/discrete_depth_distortion_model.h) as static methods:
```cpp
class RTABMAP_EXP DiscreteDepthDistortionModel
{
public:
   // returns all divisors of num
   static std::set<size_t> getDivisors(const size_t &num);
   
   // returns divisor from divisors closest to ref
   static size_t getClosestToRef(const std::set<size_t> &divisors, const double &ref);
   
   // sets bin_width and bin_height to appropriate values
   static void getBinSize(const size_t &width, const size_t &height, size_t &bin_width, size_t &bin_height);

public:
    DiscreteDepthDistortionModel() : ...
```

Then in DepthCalibrationDialog.cpp, we would use it like this:
```cpp
size_t bin_width, bin_height;
clams::DiscreteDepthDistortionModel::getBinSize(imageSize.width, imageSize.height, bin_width, bin_height);
_model = new clams::DiscreteDepthDistortionModel(imageSize.width, imageSize.height, bin_width, bin_height);
```

After merging, I may add `bin_depth`, `smoothing` parameters to dialog, maybe `bin_width` and `bin_height` too to set reference bin size for your code.

cheers,
Mathieu

T0ny0:
I moved new methods to clams namespace. For `512 * 424` image bin size will be `8 * 8`. For `640 * 480`: `8 * 6` (as it was before by default). Good idea to add this parameters to dialog. `getBinSize` can be used to set default values of bin_width and bin_height in dialog. And there should be a tooltip that says that bin size must be multiple of image size.

matlabbe:
Thx for the changes. I'll try to add your suggestion for the default values.

matlabbe:
Distortion parameters are now added to dialog for convenience: https://github.com/introlab/rtabmap/commit/414e3555a546afef9b6eefffc2579e1cfc3b3e17

Issue #188, opened 2017-05-05
Support for absolute position/ attitude measurements


jgoppert:
I would like to add links to the graph-slam algorithm in rtabmap for measurements of absolute position and attitude. For instance, GPS, or known landmarks in an environment. Please let me know if this is already implemented. If it hasn't been implemented I would be happy to add it and feel capable of doing so with a few pointers of where to add it in the code. I presume I would need to add a new type of sensor data.

matlabbe:
Hi,

For attitude, this could be somewhat included in odometry values. For absolute position, there is currently no way in rtabmap to add them (as fixed constraints in the graph for example). With g2o graph optimization approach selected, there could be a way to do this. See [OptimizerG2O::optimize()](https://github.com/introlab/rtabmap/blob/6abed48af69ad46db679f5c281805bc448b5213e/corelib/src/OptimizerG2O.cpp#L141). For example, we may want to add new g2o::VertexSE3 to graph and set them as fixed [like](https://github.com/introlab/rtabmap/blob/6abed48af69ad46db679f5c281805bc448b5213e/corelib/src/OptimizerG2O.cpp#L239) the root node. Then add g2o::EdgeSE3 between the corresponding node of the map's graph and this new absolute node (with some covariance).

[SensorData](https://github.com/introlab/rtabmap/blob/master/corelib/include/rtabmap/core/SensorData.h) could be extended with GPS coordinate (relative to start or geo-referenced) + GPS covariance.

To pass these values to g2o, that could be done in the [Memory::getMetricConstraints()](https://github.com/introlab/rtabmap/blob/1b71a96a3db64b019f0f26c95d430ccf47958f6a/corelib/src/Rtabmap.cpp#L3080), which could get GPS pose like the odom pose [here](https://github.com/introlab/rtabmap/blob/e415dd8504cb48f4085fb23ff4e8d75161cb5022/corelib/src/Memory.cpp#L4067), then we call optimize() [here](https://github.com/introlab/rtabmap/blob/1b71a96a3db64b019f0f26c95d430ccf47958f6a/corelib/src/Rtabmap.cpp#L3144) with added absolute GPS_poses + GPS_covariances.

Before going into code, you could export the rtabmap's graph in g2o format (see file->Export poses...), add manually these absolute constraints to g2o file, then use `g2o_viewer` tool to optimize this file and see results (similar to the end of this [tutorial](https://github.com/introlab/rtabmap/wiki/Robust-Graph-Optimization)).

[...]

PR #196, opened 2017-05-17
WIP: Add edge prior to g2o.

jgoppert:
@matlabbe Let me know what you think. This is my first stab at this. Will test it out soon in gazebo.

Closes: https://github.com/introlab/rtabmap/issues/188


jgoppert:
Currently building and I can run it, but getting a g2o except due to the offset you have to set. Shouldn't be too bad tracking it down:

```bash
[ WARN] (2017-05-17 19:46:46.792) OptimizerG2O.cpp:388::optimize() adding edge prior
terminate called after throwing an instance of 'std::bad_typeid'
  what():  std::bad_typeid
```

jgoppert:
This runs now. The export does not have the new links, but I suspect this is because the export to g2o just recreates the links and edges from the poses.

matlabbe:
The export is done in [OptimizerG2O::saveGraph()](https://github.com/introlab/rtabmap/blob/6abed48af69ad46db679f5c281805bc448b5213e/corelib/src/OptimizerG2O.cpp#L918), so yes it will not export the new links unless we add them.

I will have some time this weekend to try this code. Meanwhile, what actually means g2o:: EdgeSE3Prior? Should it be the prior of a pose? Like saying that pose should match this GPS value (measurement) with this error (information matrix). If so, that could be done in the [poses loop](https://github.com/dronecrew/rtabmap/blob/a2a8638f2810455da6f9e0850f1722ee3395de65/corelib/src/OptimizerG2O.cpp#L219) instead of the [edgeConstraints loop](https://github.com/dronecrew/rtabmap/blob/a2a8638f2810455da6f9e0850f1722ee3395de65/corelib/src/OptimizerG2O.cpp#L254). The information matrix between two links "42" -> "43" may not have the same meaning than information matrix set to "43" pose prior. Note also that not all poses may have a prior (e.g., marker/gps pose).  

cheers,
Mathieu

jgoppert:
@matlabbe Yes, it is the prior of a pose with the given information matrix, so it would make more sense in the pose loop, I will move it and add it to the export so we can debug more easily.

>  Note also that not all poses may have a prior (e.g., marker/gps pose).

I agree. This approach isn't actually creating a pose for the known  landmarks since it would be a waste of states since you effectively have infinite information. Instead it is just recording the vehicle pose at a fixed interval. So the known  landmark measurement is just shrinking the covariance of the vehicle pose before it records it.

Eventually I would like to add estimation of some unknown visual landmarks as well. These would have priors, and to add them I envision adding a binary SE3 edge.

jgoppert:
I tried moving the edge prior to the pose loop, but the issue that we don't have the separate information stored for that yet, so we would have to figure out how best to do that.

matlabbe:
I added members `globalPose` and `globalPoseCovariance` to `SensorData` [here](https://github.com/introlab/rtabmap/blob/e8af8f7792e1439aba9787e25ebd62dff0076027/corelib/include/rtabmap/core/SensorData.h#L266-L267). These are translated as unary link [here](https://github.com/introlab/rtabmap/blob/e8af8f7792e1439aba9787e25ebd62dff0076027/corelib/src/Memory.cpp#L3859-L3863) when a signature is created. This link will appear in `OptimizerG2O::optimize()` [here](https://github.com/introlab/rtabmap/blob/e8af8f7792e1439aba9787e25ebd62dff0076027/corelib/src/OptimizerG2O.cpp#L264-L308).

Using `Link` class to pass the global pose and covariance to g2o was the easiest way to do this (I think). I had to merge with `devel` branch instead of `master` because I was working from the past two weeks on changes affecting that pull request.

For now I tested with KITTI dataset by setting (hard-coded) ground truth poses as the global pose with some covariance. However, I'm not fixed on how to set appropriate covariance/information matrix for the global pose without causing some strange graph optimization errors.

matlabbe:
I [updated](https://github.com/introlab/rtabmap_ros/commit/b6b5a1795ebc926c4e317bcae19591cd5795871b) rtabmap_ros with a new topic called `global_pose`. Sending a pose with covariance on this topic, and the it will be saved in the created node (so that g2o can optimize with it). Not yet tested though, only in `devel` branch of rtabmap_ros.

jgoppert:
Awesome, thanks! Will test it out soon and let you know how it is working for me.

jgoppert:
@matlabbe

Currently it looks like there is a bug in the covariance copy for odom. Any idea what is going on, this is what I got from gdb.

```c++
 covariance = cv::Mat(6,6,CV_64FC1, (void*)odomMsg->twist.covariance.data()).clone();
```
```bash
p *(odomMsg->twist.covariance.data())@36
$21 = {1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1}
```

```bash
p *(covariance->data)@36
$17 = "\000\000\000\000\000\000\360?", '\000' <repeats 27 times>
```

matlabbe:
I cannot reproduce the error. Can you echo the `global_pose` topic?
```bash
$ rostopic echo /rtabmap/global_pose
```

I tried this:
```bash
$ roslaunch freenect_launch freenect.launch depth_registration:=true
$ roslaunch rtabmap_ros rtabmap.launch args:="-d --Optimizer/Strategy 1"
$ rostopic pub -r 1 /rtabmap/global_pose geometry_msgs/PoseWithCovarianceStamped "header:
  stamp: now
  frame_id: 'camera_link'
pose:
  pose:
    position: {x: 0.0, y: 0.0, z: 2.0}
    orientation: {x: 0.0, y: 0.0, z: 0.0, w: 1.0}
  covariance: [1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1]"
```
It is not useful (global pose is always fixed), but at least it could have reproduce your bug though. I can reproduce the assert with:
```bash
$ rostopic pub -r 1 /rtabmap/global_pose geometry_msgs/PoseWithCovarianceStamped "header:
  stamp: now
  frame_id: 'camera_link'
pose:
  pose:
    position: {x: 0.0, y: 0.0, z: 2.0}
    orientation: {x: 0.0, y: 0.0, z: 0.0, w: 1.0}
  covariance: [0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1]"
```
but it is okay to assert with 0 covariance on first value.

jgoppert:
@matlabbe I'm currently not publishing global_pose, I just was trying to get the old setup working first before adding that.

```xml
    <!-- SLAM -->
    <node name="rtabmap" pkg="rtabmap_ros" type="rtabmap" output="$(arg rtabmap_output)" args="$(arg rtabmap_args)" required="true" launch-prefix="$(arg launch_prefix)">
      <param name="frame_id"        type="string" value="fcu"/>
 
      <remap from="rgb/image"       to="$(arg rgb_topic)"/>
      <remap from="depth/image"     to="$(arg depth_topic)"/>
      <remap from="rgb/camera_info" to="$(arg camera_info_topic)"/>
      <remap from="odom"            to="/mavros/local_position/odom"/>
      <!--Use g2o-->
      <param name="Optimizer/Strategy" type="string" value="1"/>
      
      <param name="Kp/DetectorStrategy"    type="string" value="6"/>
      <param name="RGBD/LocalRadius"  type="string" value="10.0"/>
      <param name="RGBD/CreateOccupancyGrid"  type="bool" value="true"/>

      <!--localization mode-->
      <param name="Mem/InitWMWithAllNodes" type="string" value="$(arg localization)"/> 
      <param     if="$(arg localization)" name="Mem/IncrementalMemory" type="string" value="false"/>
      <param unless="$(arg localization)" name="Mem/IncrementalMemory" type="string" value="true"/>
      
    </node>
```

```bash
rostopic  echo /mavros/local_position/odom
...
header: 
  seq: 20068
  stamp: 
    secs: 672
    nsecs: 699593226
  frame_id: local_origin
child_frame_id: fcu
pose: 
  pose: 
    position: 
      x: -0.000769741251133
      y: -0.000116091105156
      z: 0.00248700310476
    orientation: 
      x: 0.0184923064716
      y: -6.77194488448e-05
      z: -0.700763868829
      w: -0.713153569786
  covariance: [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
twist: 
  twist: 
    linear: 
      x: -0.00504005171374
      y: -0.000104776311281
      z: 0.0227491119898
    angular: 
      x: 0.00145531632006
      y: -0.000326997600496
      z: 0.00246357638389
  covariance: [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
---
```

## Software version for test

I see that you have some new commits, trying those now, but this is how I got the error.

```versions:

rtabmap

commit 6f9ac4b88b67889e5465a69b44748560618a821c
Author: matlabbe <matlabbe@gmail.com>
Date:   Tue May 23 17:30:23 2017 -0400

    fixed some compilation warnings

rtabmap_ros

commit b6b5a1795ebc926c4e317bcae19591cd5795871b
Author: matlabbe <matlabbe@gmail.com>
Date:   Tue May 23 17:47:56 2017 -0400

    rtabmap: Added global_pose input topic (can be used to add GPS/global localization values for graph optimization using g2o)

```

matlabbe:
Can you try with this commit? https://github.com/introlab/rtabmap_ros/commit/252d5ed355b4afdf539347ec0990d222b5e1c2ae

I forgot to initialize globalPose_ member with time to zero. Not sure why on my system it was correctly initialized to 0 without explicitly set it. Having non-zero would make rtabmap assume that global_pose is set while it is not.

matlabbe:
Edit: reproduced the bug with robot_mapping.bag demo. Working on it...

matlabbe:
That commit should fix the problem: https://github.com/introlab/rtabmap_ros/commit/04bbb59870a9b9e0669cd099a28e2e5396042341
The covariance was wrongly set to all ones instead of identity (creating not valid information matrix when inverted).

jgoppert:
@matlabbe Yeah, that was it. I stared at that long enough, surprised I didn't find it, thanks!

nes123:
really great work. 
Is there a working example or a document that can give an intro on how to use this addition ?

paul-michalik:
@jgoppert @matlabbe I'd second the request of @nes123 ... Is there any sort of document or example which would describe the solution in more detail? I am trying to add absolute pose constraints in our ORB-SLAM2 based experiment and I do struggle with parts of it.

matlabbe:
The edge prior feature has never been really tested (very experimental!), so this is why there is no documentation/example about it. The code has been modified since this pull request. To enable it, we should set `Optimizer/PriorsIgnored=false`. Make sure to use `Optimizer/Strategy=1` (g2o). To send a prior to rtabmap (ros node), this can be done with `/rtabmap/global_pose` topic (which is a [geometry_msgs::PoseWithCovarianceStamped](http://docs.ros.org/api/geometry_msgs/html/msg/PoseWithCovarianceStamped.html)). This will add the prior to next node created as a self-referenced link. [OptimizerG2O](https://github.com/introlab/rtabmap/blob/master/corelib/src/OptimizerG2O.cpp) will detect them and add the EdgeSE3Prior or EdgeSE2Prior with pose and covariance for graph optimization. 

The feature was meant to be used with a global localization system (like a VICON) giving an external "true" pose of the camera when available. In commit above https://github.com/introlab/rtabmap_ros/commit/a735069ba9a94b39c1f058b76eac1ac2a3f1419c, I added three files as example. Look at the header of test_prior.launch for more info.

With Prior (2.1 cm error, I forgot but I think I launched test_prior_tf_to_pose.py with small covariance: 'test_prior_tf_to_pose.py _cov:=0.001'): 
screenshot from 2018-02-09 16-50-29

Without Prior (3.4 cm error):
screenshot from 2018-02-09 16-50-53



cheers,
Mathieu

paul-michalik:
@matlabbe Are you sure the screenshots aren't identical? I don't see a difference... How do you determine the covariance in the measurements for a `EdgeSE3Prior`?

paul-michalik:
Thanks, will try to look at it. Any idea how to determine good covariances for the `EdgeSE3Prior` measurements in case they aren't known I advance? We have made some experiments (in other project which uses g2o too) with estimated values but the outcome wasn't good...

matlabbe:
See bottom left, the RMSE is different. At this low error, it is indeed difficult to see the difference. I may have zoomed more, the bottom of the path was following more the ground truth using prior.

Note that this case is just for example, as if you have a VICON, use it directly as odometry instead of visual odometry. Mixing visual odometry and VICON like that would be more useful if at some point we cannot have VICON estimate, because the camera goes outside of the tracking area.

I don't know from which kind of system you want to inject priors, but covariance of priors may be set to 1 to 0.01 if they are less accurate than vo. If more acurate than vo, then set it below 0.0001. Make sure TF between base link, camera link and tracking link is accurate. In this example VICON was tracking kinect link (used as base link) and there is a transform between kinect and the actual camera frame.

If you want to do sensor fusion instead, look at the robot_localization package.

paul-michalik:
I want to fuse VO with GPS measurements. Probably only the position (latitude, longitude) will be available, maybe some error estimations. The intention is to correct the scale errors and to put the local frame of reference in relation to the GPS coordinates. I'm considering either prior injection of position information or binary constraints between poses for which GPS measurements are known, maybe both. Any experience with this?

matlabbe:
I did try integrating GPS to 6DOF mapping, without success. Here is my test: https://github.com/introlab/rtabmap/blob/400952b3276217e2b7c9f0b29bc0ee99bcd97f13/corelib/src/Memory.cpp#L4034-L4052

The idea was to set very high variance on degrees of freedom not used by GPS (z, roll, pitch, yaw), and variance to x,y based on GPS error (e.g., 5 meters of error), however the resulting graph was really poorly optimized. I didn't have time to look more into that though. g2o has examples with GPS but they are only in 3DoF if I remember.

TSC21:
@matlabbe friendly ping

matlabbe:
The exportation of prior is not implemented. In the file you have you may change all EDGE_SE3:QUAT that have same ID to EDGE_SE3_PRIOR (keeping only one id). 

The priors above seem to have identity information matrix, which may not have a big effect on the resulting optimization.

PR #231, opened 2017-08-31
KAZE - fix loading custom feature params

mlong-xii:
KAZE feature detector does not load custom feature params (min/max depth, ROI, etc).


matlabbe:
Good catch! thx!

mlong-xii:
Definitely. Glad to help.

PR #273, opened 2018-03-28
Fix laserscan from depth

mlong-xii:
Mathieu, unless I'm missing something, you intended to change this to isEmpty() correct?


matlabbe:
Yes, thanks to find out! The error was introduced after adding the recent LaserScan class.

mlong-xii:
No problem. Glad to help.

PR #284, opened 2018-05-27
PreferencesDialog: Add forgotten include

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>


matlabbe:
Is it because there is a compilation error on a specific platform (qt version)? Normally, QButtonGroup should be included by the generated ui_preferencesDialog.h.

patrickelectric:
I'm having a huge number of problem with qt5.11
It was necessary to change the cmakefile also, I can send a .diff you want

PR #289, opened 2018-06-20
fix compilation issue

facontidavide:
Hi,

I got a compilation error due to the static member (undefined reference).
I am compiling on Ubuntu 16.04.

This change fixed the problem.

Cheers


matlabbe:
What is the version of your gcc for info as I cannot reproduce the compilation error? The way the static member is initialized was taken from the octomap examples: https://github.com/OctoMap/octomap/blob/aef5c4593712ed81230c6d44ef0091a8de45bf10/octomap/include/octomap/ColorOcTree.h#L198

matlabbe:
This bug has been reproduced in Debug mode in issue #339. The commit above added the missing definition of the static member in the cpp.

Issue #499, opened 2020-01-29
ZED SDK 3.0 breaks the compilation


lazydroid:
Recently (last week) released SDK for ZED cameras from Stereolabs breaks the compilation because of newly introduced types and changes in the namespace. SDK 2.8.* works fine, but does not support the new ZED2 camera.

lazydroid:
I have made the changes to ZED related file that allow RTABMap to get compiled with SDK 3.0 and it works with ZED 2. Attaching the file if anyone is interested.

[CameraStereoZed.cpp.zip](https://github.com/introlab/rtabmap/files/4137758/CameraStereoZed.cpp.zip)

matlabbe:
See pull request #501 and test it. I'll close this issue in the meantime, feel free to comment if the pull request is not working for you.

lazydroid:
Your code fails to compile because the arguments of `RuntimeParameters` have changed :

```
rtabmap/corelib/src/camera/CameraStereoZed.cpp:545:121: error: no matching function for call to ‘sl::RuntimeParameters::RuntimeParameters(sl::SENSING_MODE, bool, bool, sl::REFERENCE_FRAME)’
```

matlabbe:
The commit above should fix the problem, thx for testing it!

lazydroid:
It compiles and works now, thank you very much for your support!!

PR #501, opened 2020-01-31
Compatibility support for ZED SDK v3.0

Myzhar:
Add support for Stereolabs ZED SDK v3.0 
Guaranteed compatibility with older SDK v2.x


matlabbe:
Great! Thx a lot!
It should fix issue #499 as well.

PR #551, opened 2020-05-15
Added Kinect for Azure menu option

tkircher:
In anticipation of Kinect for Azure device support


matlabbe:
Great! Thank you. I should be able to test it soon.

PR #552, opened 2020-05-20
Preliminary K4A device support

tkircher:


tkircher:
Strange that github thinks all the previous commits are part of the PR. Diffs look good though.

PR #618, opened 2020-10-24
[Windows] Fix VTK\QT and OpenCV discovery

seanyen:
This pull request is to fix the case where it is building with `WITH_QT=OFF` (which wouldn't find VTK consequently) and OpenCV w/o `NONFREE` module built (tested with OpenCV `4.3`).


matlabbe:
The old define of the opencv part of your version was indeed wrong, it has been updated in this commit:
https://github.com/introlab/rtabmap/commit/a4da1e14b4e6ab9e9bd69a9875fb679a58d51f74#diff-80566d67599fb1bfd180da18de5c6c456f44c71648d5263346c7d1efbd4972ba
From 4.4.0, SIFT is not in NONFREE module anymore. Same for 3.4.11, which makes a more complicated define...

VTK could be included by PCL indirectly. DISABLE_VTK definition is from [PCL definitions normally](https://github.com/PointCloudLibrary/pcl/blob/bc1cb2c2aba134c3436e56e294252fe7cb43b54d/PCLConfig.cmake.in#L339), if PCL has not been built with VTK. VTK is used in PCL for some filtering functions (like mesh decimation):
https://github.com/introlab/rtabmap/blob/bda18bc51769d2ddf7001653d75a688789b05538/corelib/src/util3d_surface.cpp#L62

matlabbe:
Can you show the cmake log status on your build? If PCL is not build with VTK, there should be a `-DDISABLE_VTK` in the `PCL_DEFINITIONS` like this:
```
--   PCL_DEFINITIONS =  ;-DDISABLE_VTK;-DDISABLE_ENSENSO;-DDISABLE_DAVIDSDK;-DDISABLE_DSSDK;-DDISABLE_PCAP;-DDISABLE_PNG;-DDISABLE_LIBUSB_1_0;-DFLANN_STATIC;-Dqh_QHpointer;-DDISABLE_RSSDK
```
Not that DISABLE_VTK should not be set if PCL is built with VTK.

seanyen:
@matlabbe I am using the PCL port (without `VTK` features) from the Vcpkg to test the build. And here is the related outputs from CMake:
```
-- Info :
--   Version : 0.20.3
--   CMAKE_INSTALL_PREFIX = D:/packages/rtabmap_x64-windows
--   CMAKE_BUILD_TYPE =     Release
--   CMAKE_INSTALL_LIBDIR = lib
--   BUILD_APP =            ON
--   BUILD_TOOLS =          ON
--   BUILD_EXAMPLES =       ON
--   BUILD_AS_BUNDLE =      OFF
--   CMAKE_CXX_FLAGS =  /nologo /DWIN32 /D_WINDOWS /W3 /utf-8 /GR /EHsc /MP  /MP16 -openmp
--   FLANN_KDTREE_MEM_OPT = OFF
--   PCL_DEFINITIONS = -D__SSSE3__;-D__SSE2__;-D__SSE__;-DBOOST_ALL_NO_LIB;-DDISABLE_PNG;-DDISABLE_QHULL
--   PCL_VERSION = 1.11.1
```

matlabbe:
I didn't try PCL 1.11.1 on Windows lately, but it seems it is built with VTK as there is no `-DDISABLE_VTK` in `PCL_DEFINITIONS`

windelbouwman:
See also: https://github.com/introlab/rtabmap/issues/795

I think relying on the `DISABLE_VTK` is tricky.

- Reason 1: `#ifndef DISABLE_VTK` is a double negation, I like `#ifdef WITH_VTK` better
- Reason 2: When the define is absent, it is assumed we do use VTK, which might not be the case.

However, PCL defines multiple more defines starting with `DISABLE_`, what happens when those defines are removed?

matlabbe:
Didn't notice before, but you should make pull request on master branch, devel is pretty old. That should fix OpenCV SIFt problem. For DISABLE_VTK, see this pull request instead: CV_MAJOR_VERSION > 4

PR #659, opened 2020-12-10
superpoint, fixing slow convert after cnn

42Max42:
I noticed that the superpoint in GPU was not as fast as expected, leading to over 1s keypoint detection time, even when using gpu.
I found that when using the gpu the tensors were still stored in gpu. I made a little modification and it helps to reduce a lot the keypoint detection time.
also related to: https://github.com/introlab/rtabmap/issues/633#issuecomment-727099760


CMakeLists.txt
```diff
@@ -207,7 +209,7 @@ ENDIF()
 set(RTABMAP_QT_VERSION AUTO CACHE STRING "Force a specific Qt version.")
 set_property(CACHE RTABMAP_QT_VERSION PROPERTY STRINGS AUTO 4 5)
 
-FIND_PACKAGE(OpenCV REQUIRED QUIET)
+FIND_PACKAGE(OpenCV 3 REQUIRED QUIET)
```
matlabbe: Remove hard-coded 3. IF you have OpenCV 2 or OpenCV4 also installed, use `-DOpenCV_DIR=/path/to/OpenCVConfig.cmake/directory` when doing cmake to select your opencv 3 installation.

CMakeLists.txt
```diff
@@ -4,7 +4,9 @@ cmake_minimum_required(VERSION 2.8.5)
 PROJECT( RTABMap )
 SET(PROJECT_PREFIX rtabmap)
 
-# Catkin doesn't support multiarch library path, 
+set(CMAKE_CXX_STANDARD 14)
```
matlabbe: Remove this, it is already added with Torch when found here: https://github.com/introlab/rtabmap/blob/da8e76ffed836fad3319716e5e15112a59e4e4f5/CMakeLists.txt#L607-L617

matlabbe (changes requested):
Thx for changes, but see my 2 comments for minor issues.

matlabbe:
Great, thank you!

PR #713, opened 2021-05-04
turn off building with ZED

fiorano10:
No reason to build with ZED


matlabbe:
This option is set to true by default as there is not know building issue to include it by default is zed sdk is installed. Otherwise, you are free to do `cmake -DWITH_ZED=OFF` to not build with it.

In general, options that are OFF by default it is because there were some compatibility issues depending on the system or with other dependencies.

PR #775, opened 2021-10-22
Add ceiling filter to rtabmap-export

iportolesrb:
Added a parameter in rtabmap-export to filter ceiling if wanted.

### DIFF
--- a/tools/Export/main.cpp
+++ b/tools/Export/main.cpp
@@ -115,6 +115,8 @@ void showUsage()
 			"    --ymax #              Maximum range on Y axis to keep nodes to export.\n"
 			"    --zmin #              Minimum range on Z axis to keep nodes to export.\n"
 			"    --zmax #              Maximum range on Z axis to keep nodes to export.\n"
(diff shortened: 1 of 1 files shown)


tools/Export/main.cpp
```diff
@@ -115,6 +115,8 @@ void showUsage()
 			"    --ymax #              Maximum range on Y axis to keep nodes to export.\n"
 			"    --zmin #              Minimum range on Z axis to keep nodes to export.\n"
 			"    --zmax #              Maximum range on Z axis to keep nodes to export.\n"
+			"    --filter_ceiling #    Filter points over a custom height (dafault 2 m, 0=disabled).\n"
```
Timple, line 118: Typo :slightly_smiling_face:

matlabbe: See the final version here:
https://github.com/introlab/rtabmap/blob/ddd5eb5a41d08784db4211ffec6f5eb88e4afac8/tools/Export/main.cpp#L118-L119

Timple: Whups, sorry.

I was quickly browsing to see what changed in the new version. Should have looked at the end-result :slightly_smiling_face:

PR #810, opened 2022-01-07
Add Boost serialization package for libpointmatcher case

MarvinStuede:
Setup:

- Ubuntu 20.04 with ROS noetic
- Build with GTSAM, G2o and libpointmatcher
- Boost Version 1.71

During Cmake Step, I had warnings that the Boost::serialization module was not found. This resulted in a "links to target "Boost::serialization" but the target was not found" error during compilation. Adding the serialization module to CMakelists resolved this.


matlabbe:
Thx!

PR #811, opened 2022-01-11
Add libpointmatcher as dependency

Timple:
As per: https://github.com/introlab/rtabmap_ros/pull/700


matlabbe:
Included in https://github.com/introlab/rtabmap/pull/812

PR #815, opened 2022-01-24
Allow negative disparity in dialog for stereo matchers

benknight135:
Some stereo cameras need to allow negative disparity. Depth will still be possitive values but negative disparity occurs when depth value is behind cross over point of toed in cameras. StereoBM & StereoSGBM classes allow negative values but dialog limits between 0 and 999999. This PR changes the minimum to -999999 in the dialog for StereoBM and StereoSGBM.


benknight135:
I applogise I need to close this PR so I can continue to use my own fork of the repository. I will re-open with this change in a branch.

PR #816, opened 2022-01-24
fix missing curl install in focal dockerfile

benknight135:
Fix K4A binary download step in focal dockerfile failure due to missing curl install. Curl could be added earlier in the dockerfile but thought it best to keep the scope of the fix local to the install of k4a so as not to affect eailier build steps cache.


matlabbe:
Indeed, not sure why it worked the first time, and that GitHub actions (creating the docker images) are not failing. Just tried creating the docker from scratch and we do have `"/bin/sh: 1: curl: not found"` error. Thx for the check!

PR #817, opened 2022-01-24
Allow negative disparity in dialog for stereo matchers

benknight135:
Some stereo cameras need to allow negative disparity. Depth will still be possitive values but negative disparity occurs when depth value is behind cross over point of toed in cameras. StereoBM & StereoSGBM classes allow negative values but dialog limits between 0 and 999999. This PR changes the minimum to -999999 in the dialog for StereoBM and StereoSGBM.


matlabbe:
For reference:

In [StereoBM](https://docs.opencv.org/4.x/d9/dba/classcv_1_1StereoBM.html) doc, they don't explain what is minDisparity, just the following is refering it:

> numDisparities the  disparity search range. For each pixel algorithm will find the best  disparity from 0 (default minimum disparity) to numDisparities. The  search range can then be shifted by changing the minimum disparity.

It is more clear in [StereoSGBM](https://docs.opencv.org/4.x/d2/d85/classcv_1_1StereoSGBM.html#adb7a50ef5f200ad9559e9b0e976cfa59) doc:

> minDisparity Minimum  possible disparity value. Normally, it is zero but sometimes  rectification algorithms can shift images, so this parameter needs to be  adjusted accordingly.
> numDisparities Maximum disparity  minus minimum disparity. The value is always greater than zero. In the  current implementation, this parameter must be divisible by 16.

Thx

PR #871, opened 2022-05-28
 fixing missing DISABLE_VTK on recent PCL versions

matlabbe:


windelbouwman:
This surely would fix building with newer version of the PCL library, so this is a good change.

Still, I believe the code could be improved even further by not relying on the `DISABLE_VTK` define at all in the rtabmap code, since that define is coming from the PCL library (it belongs to PCL, not rtabmap code), so it might change in the future. Instead, it might be good to create an own define, something like `RTABMAP_WITH_VTK`, to indicate that rtabmap will be build using VTK.

Anyway, good change, this will solve compilation issues!

matlabbe:
I agree, this was a quick fix (quite busy right now). Defining RTABMAP_WITH_VTK for c++ based on VTK_FOUND cmake variable would be better. 

I think it would work in thoses cases:
* PCL built with VTK, rtabmap built with Qt
* PCL built with VTK, rtabmap not built with Qt
* PCL not built with VTK, rtabmap built without Qt
* PCL not built with VTK, rtabmap built with Qt: this is the case I thought DISABLE_VTK was needed, but this should correctly fail by cmake because pcl visualization component won't be found.

PR #875, opened 2022-06-27
Correct Jacobian for XYFactor

kurtgeebelen:
The Jacobian for the XYFactor was incorrect. This happens to be the example for the Unary factor on gtsam documentation, where the correct Jacobian is given:
https://gtsam.org/tutorials/intro.html#:~:text=In%20GTSAM%2C%20you%20can%20create,%7D%20%3D%20f%20(%20q%20)%20where

### DIFF
--- a/corelib/src/optimizer/gtsam/XYFactor.h
+++ b/corelib/src/optimizer/gtsam/XYFactor.h
@@ -45,8 +45,8 @@ public:
 
     // note that use boost optional like a pointer
     // only calculate jacobian matrix when non-null pointer exists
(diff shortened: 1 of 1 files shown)


kurtgeebelen:
FYI: if you want, I can additionally make a pull request where you can (optionally) provide a position of the GPS receiver on your robot, and account for that offset (since the receiver may not always be in the origin of your robot).

kurtgeebelen:
Hmm, I know see that you have made it a template class (before it was only for Pose2). In that case my solution does not work anymore (but the Jacobian remains incorrect I think for Pose2).

kurtgeebelen:
Ok, with the templated class this does not work. I'll propose another fix.

PR #955, opened 2023-01-14
OpenCV 4.7 moved aruco, and has breaking changes

cnSchwarzer:
OpenCV 4.7 moved aruco, and has breaking changes. This PR adds situations for it.


matlabbe:
Great thx!

PR #957, opened 2023-01-20
fix bigobj under MSVC, fix yaml-cpp link.

cnSchwarzer:
- MSVC needs `/bigobj` to compile
- The original cmake file failed to link `yaml-cpp`, as the `find_package` doesn't set `YAML_CPP_INCLUDE_DIRS` and `YAML_CPP_LIBRARIES` (under `vcpkg`, at least), which failed to link. This patch link `yaml-cpp` directly.


matlabbe:
Thanks!

matlabbe:
I checked the [old msvc build](https://ci.appveyor.com/project/matlabbe/rtabmap/builds/45956483), and `/bigobj` seems not used and worked. I remember though on some computer config it is required. Having the flag doesn't seem to cause build issue, I'll keep it.

For yaml-cpp, maybe it is because on how vcpkg build its own version of yaml-cpp (without old cmake include/library variables). The windows build seems [now](https://ci.appveyor.com/project/matlabbe/rtabmap/builds/45994816) including two times the yaml-cpp library:
```
"C:\Program Files\yaml-cpp\lib\libyaml-cppmd.lib" "C:\Program Files\yaml-cpp\lib\libyaml-cppmd.lib"
```
This doesn't affect the build either, so lets keep it for vcpkg.

PR #1018, opened 2023-04-18
F2M: use guess when transformation estimation failed

guoqingzh:
This make the F2M recover quicker from the transformation estimation failure and reduce the amount of countdown for reset

### DIFF
--- a/corelib/src/Rtabmap.cpp
+++ b/corelib/src/Rtabmap.cpp
@@ -3169,7 +3169,7 @@ bool Rtabmap::process(
 					UWARN("Optimization failed, rejecting localization!");
 					rejectLocalization = true;
 				}
-				else if(_optimizationMaxError > 0.0f)
+				else
 				{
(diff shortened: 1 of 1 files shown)


matlabbe:
I am not sure to understand what it improves. Here what the current code is doing:

If a guess is provided, estimate first with the guess (local feature matching). If estimated transform is null (failed), try again without guess to force a global feature matching (in case the guess is wrong). If one of this two steps success, proceed and return the new estimated pose, otherwise return failure.

From what I understand from the modification of this PR:

If a guess is provided, estimate first with the guess (local feature matching). If estimated transform is null (failed), set `transform` to non null to exit the transformation estimation loop (thus skipping global feature matching step), then exit odom update with failure.

So overall, this makes F2M never do global feature matching if there is always an input guess. If you want to reduce the amount of countdown to reset, set `Odom/ResetCountdown` to 1.

PR #1033, opened 2023-05-14
Fixed build with gtsam 4.3.0

matlabbe:


wakoko79:
@matlabbe cc17ebe92b8276d36d001b27d3b0d67cc845edfc breaks gtsam 4.3.0 support.
At line 704 of rtabmap/corelib/src/optimizer/OptimizerGTSAM.cpp : gtsam::Values::iterator is used, but it only accessible with gtsam 4.1.1

matlabbe:
Should be fixed in https://github.com/introlab/rtabmap/commit/063ca9c8eeab7197e51c39c6f27b8cf9541a5849, thx for reporting!

PR #1034, opened 2023-05-17
Update DepthAI Integration

borongyuan:
Add option to enable IR laser dot projector (active stereo) and IR illumination LED (for night-vision).
The default value for laser dot projector is 0 (turn off). Because the projected pattern is good for depth estimation but may harm feature extraction.
It looks like there are other issues with DepthAI integration. I'll try to troubleshoot and fix them soon.

### DIFF
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -58,6 +58,7 @@ public:
 	void setOutputDepth(bool enabled, int confidence = 200);
 	void setIMUFirmwareUpdate(bool enabled);
 	void setIMUPublished(bool published);
+	void publishInterIMU(bool enabled);
 	void setLaserDotBrightness(float dotProjectormA = 0.0f);
 	void setFloodLightBrightness(float floodLightmA = 200.0f);
 
@@ -78,14 +79,15 @@ private:
 	int resolution_;
 	bool imuFirmwareUpdate_;
 	bool imuPublished_;
+	bool publishInterIMU_;
 	float dotProjectormA_;
 	float floodLightmA_;
 	std::shared_ptr<dai::Device> device_;
 	std::shared_ptr<dai::DataOutputQueue> leftQueue_;
 	std::shared_ptr<dai::DataOutputQueue> rightOrDepthQueue_;
-	std::shared_ptr<dai::DataOutputQueue> imuQueue_;
 	std::map<double, cv::Vec3f> accBuffer_;
 	std::map<double, cv::Vec3f> gyroBuffer_;
+	UMutex imuMutex_;
 #endif
 };
 
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -58,6 +58,8 @@ public:
 	void setOutputDepth(bool enabled, int confidence = 200);
 	void setIMUFirmwareUpdate(bool enabled);
 	void setIMUPublished(bool published);
+	void setLaserDotBrightness(float dotProjectormA = 0.0f);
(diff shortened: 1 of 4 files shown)


matlabbe:
Great! can't way to try a OAK Pro model in the dark!

You said

> It looks like there are other issues with DepthAI integration. I'll try to troubleshoot and fix them soon.

Should I wait to merge till they are fixed? Are issues from the depthai SDK or the driver in rtabmap?

borongyuan:
You can merge this PR first. For other issues and improvements I need some time to work on.
I'm currently noticing the following:
- sync issue, sometimes left and right frames fail to sync, IMU fail to sync
- calibration parsing issue, correct camera model may need to be computed from the calibration data using cv::stereoRectify()
- IMU local transform, IMU model and location may be different for different cameras, PCB files are being studied

I also contributed some tuning improvements to depthai-ros a few days ago, and some of them can be used here as well.

borongyuan:
- [ ] Enable IR features
- [ ] Update camera model

borongyuan:
After some research and trial and error, I found out that I cannot use cv::stereoRectify() to get the same Projection Matrix as used by the onboard Node. In fact, Project Matrix is determined by how Depth Image is aligned. By default, Depth Image is aligned to DepthAlign::RECTIFIED_RIGHT. At this time, the Projection Matrix of the left and right and depth images is consistent with the Intrinsic Matrix of the right camera. Therefore, a reasonable configuration is to align the Depth Image to DepthAlign::RECTIFIED_LEFT. In this way, the Projection Matrix of the left and right and depth images is consistent with the Intrinsic Matrix of the left camera. Now for the translation vector of Extrinsics, only the first item is non-zero, which is used to calculate the baseline. The baseline is positive and Tx of right camera is negative, which conforms to the OpenCV convention, and is also compatible with other codes of rtabmap.

In addition, dai::CameraBoardSocket::LEFT cannot be used for the alignment setting, otherwise it will cause misalienment of depth and mono images. All of the above have been verified, so that I can use OpenCV to get rectified images consistent with those obtained from the camera. And depth and mono images are perfectly aligned.

It is a little regretful that the current rectification configuration loses a certain amount of FOV. For the OAK-D Pro W I use, the HFOV is 127 for raw mono image and 96.4208 for rectified image. I will consider using raw image or change alphaScaling later.

borongyuan:
I added some advanced parameter tuning to optimize depth quality, including:
- Improving depth accuracy by enabling subpixel mode with 4 fractional bits. Especially good for long range. 4 Fractional bits is only available with DisparityWidth::DISPARITY_64.
- Reduce MinZ by enableCompanding. This does not increase computation compared to extended disparity mode. So it will not cause frame rate drop for high-resolution inputs.
- Enable median filter, which is super fast.

You can refer to https://github.com/luxonis/depthai-ros/pull/279 for more explanation.

borongyuan:
> Great! can't way to try a OAK Pro model in the dark!

For night use, I did some preliminary tests. When IrFloodLightBrightness is set to the maximum (1500mA), using 400P stereo images, it seems to work best (with relatively high MatchesRatio and InliersRatio). This is presumably because the high-sensitivity noise of high-resolution images is more likely to affect features. I'm not going to delve into this yet, but it's an interesting point, including the effects of using different features.

borongyuan:
The major plan items are finalized so they can now be merged.
For data sync, timestamps from the device's monotonic clock are used. This should avoid a lot of exceptions. Image frame's timestamp uses the time point in the middle of the exposure interval, as suggested in the DepthAI documentation. IMU data acquisition now uses the callback method. publishInterIMU_ is also supported.

matlabbe:
I tested on my computer with my OAK-D and it works (with this minor [change](https://github.com/introlab/rtabmap/commit/5b1c9e7233b7a820cea0623dbcbf5a4dc7fee521)). 

Thanks for this PR!

PR #1042, opened 2023-05-31
Support cameras with different fx and fy, enable alphaScaling for OAK cameras

### LINKED ISSUE #1036: Issue with squeeze image
Hi, I'm trying to solve the FOV issue as mentioned in https://github.com/introlab/rtabmap/pull/1034#issuecomment-1557043379.
I've tested the usage of alphaScaling. The projection matrix can be calculated by cv::getOptimalNewCameraMatrix() correctly. But these data and images don't seem to work with rtabmap. I'm still looking for the reason, but the most likely reason is that using alphaScaling makes the image squeeze. That is to say, for the camera matrix calculated by cv::getOptimalNewCameraMatrix(), fx≠fy.
Here are the parameters I get.
left: fx=329.156006 fy=474.943451 cx=613.588454 cy=393.230591 baseline=0.075324
Here are the rectified left and right images (alpha=0.0). I put a coke in front of the camera, so you can see the squeeze.
2023-05-23 13-13-16屏幕截图
2023-05-23 13-13-35屏幕截图

borongyuan:
Fix #1036 
This makes better use of the FOV of wide-angle lens.

### DIFF
--- a/corelib/src/util3d.cpp
+++ b/corelib/src/util3d.cpp
@@ -2532,7 +2532,8 @@ cv::Point3f projectDisparityTo3D(
 			c = model.right().cx() - model.left().cx();
 		}
 		float W = model.baseline()/(disparity + c);
-		return cv::Point3f((pt.x - model.left().cx())*W, (pt.y - model.left().cy())*W, model.left().fx()*W);
+		return cv::Point3f((pt.x - model.left().cx())*W,
+			(pt.y - model.left().cy())*model.left().fx()/model.left().fy()*W, model.left().fx()*W);
 	}
 	float bad_point = std::numeric_limits<float>::quiet_NaN ();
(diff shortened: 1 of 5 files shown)


corelib/src/util3d.cpp
```diff
@@ -2532,7 +2532,8 @@ cv::Point3f projectDisparityTo3D(
 			c = model.right().cx() - model.left().cx();
 		}
 		float W = model.baseline()/(disparity + c);
-		return cv::Point3f((pt.x - model.left().cx())*W, (pt.y - model.left().cy())*W, model.left().fx()*W);
+		return cv::Point3f((pt.x - model.left().cx())*W,
+			(pt.y - model.left().cy())*model.left().fx()/model.left().fy()*W, model.left().fx()*W);
```
matlabbe, line 2536: Interesting!

PR #1047, opened 2023-06-06
add depthai gftt detector

borongyuan:
Hi, I added on-device GFTT feature according to your note. I'll try to add SuperPoint later. So later for `detectFeatures_`, 0 is None, 1 is GFTT, 2 is SuperPoint. Or change it to enum afterwards.

There are some differences between DepthAI and OpenCV in the implementation of GFTT. I'm trying to reuse parameters set in other panels, which might be a little awkward. Maybe you have a better way. I tried many parameters, mainly to increase the number of features. DepthAI first uses CornerDetector to detect corner points, and then uses FeatureMaintainer to filter out some. So when `numTargetFeatures` is set to 320 by default, it may only return more than 200 points in the end. This can make rtabmap less stable. Regarding the actual `numMaxFeatures`, the [docs](https://docs.luxonis.com/projects/api/en/latest/references/cpp/#_CPPv4N3dai23RawFeatureTrackerConfig14CornerDetector14numMaxFeaturesE) say:
> Hard limit for the maximum number of features that can be detected. 0 means auto, will be set to the maximum value based on memory constraints.

So I increased the number of memory slices to 2. This does seem to increase the number of features. Now when `numTargetFeatures` is set to 1000, more than 500 points can be obtained in most scenes. On-board detection takes about 7ms. But don't increase the number of shaves to 2, it will cause the detection time to increase to more than 30ms. I have no idea why. MotionEstimator is disabled. But we can also try to use  on-device optical flow later. Although there doesn't seem to be an interface for setting optical flow data. Return only tracked points may improve stability, or reduce host computation.

### DIFF
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -63,7 +63,7 @@ public:
 	void setLaserDotBrightness(float dotProjectormA = 0.0f);
 	void setFloodLightBrightness(float floodLightmA = 200.0f);
 	void setDetectFeatures(int detectFeatures = 0);
-	void setGFTTDetector(bool useHarrisDetector, double minDistance = 7.0f);
+	void setGFTTDetector(bool useHarrisDetector, double minDistance = 7.0f, int numTargetFeatures = 1000);
 
 	virtual bool init(const std::string & calibrationFolder = ".", const std::string & cameraName = "");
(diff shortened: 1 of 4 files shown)


matlabbe:
This is pretty cool, odometry can be processed under 30 ms when computing the features onboard (OAK-D). I also prefer re-using parameters from other panels in the settings, to make it easier to compare on-device versus host. I'll merge it! Thanks again!

PR #1068, opened 2023-06-24
msckf_vio now supports c++14

borongyuan:
C++14 is supported since https://github.com/KumarRobotics/msckf_vio/commit/12105549ad9194289827e063e9465589a43eef19.
Try this patch.
https://gist.github.com/borongyuan/90f10aa6969ebccdba2b19ce0cfa77ea


borongyuan:
MSCKF VIO is working. But there are some problems with the coordinate system. I'll solve this part later.

PR #1072, opened 2023-06-27
Remove dependency on qt_gui_cpp.

clalancette:
There is no dependency in this package on qt_gui_cpp. Instead, just use a dependency on qtbase5-dev


matlabbe:
This was a hold hack to get qt4 or qt5 dependencies based on ros distro. I think all ros distros are qt5 now, so it can be good to update it. Thx for the PR.

clalancette:
Thanks for merging!

PR #1091, opened 2023-07-11
Motion estimation: use MultiNoncentralAbsolutePoseSacProblem to estim…

guoqingzh:
…ate camera pose


matlabbe:
Merged, thx for the PR. To try multi approach, set `Vis/PnPSamplingPolicy` to 2.

PR #1094, opened 2023-07-14
OAK-D PoE Update

borongyuan:
I have received OAK-D Pro W PoE. Its IMU local transform is easy to determine, it has the same orientation as the USB model, but with a different offset. What is more troublesome is the communication bandwidth. I couldn't get enough camera frame rate and IMU frame rate with the previous configuration. The official documentation says a lot about this. I'll have to see what tuning needs to be done.
In addition, the previous USB models equipped with BMI270 also have [IMU frame rate problems](https://github.com/luxonis/depthai-ros/issues/145). Certain algorithms may be affected.

### DIFF
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -49,7 +49,7 @@ public:
 
 public:
 	CameraDepthAI(
-			const std::string & deviceSerial = "",
+			const std::string & mxidOrName = "",
 			int resolution = 1, // 0=720p, 1=800p, 2=400p
 			float imageRate=0.0f,
 			const Transform & localTransform = Transform::getIdentity());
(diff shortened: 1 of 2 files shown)


borongyuan:
VideoEncoder is now used automatically on PoE devices for compressed image transport. Only 8-bit camera images or disparity images can be encoded. So Subpixel is not supported. This reduces long range depth accuracy and causes layering effects as shown in https://github.com/luxonis/depthai/issues/184 and [Stereo subpixel effect on layering](https://docs.luxonis.com/projects/api/en/latest/tutorials/configuring-stereo-depth/#stereo-subpixel-effect-on-layering). So for PoE devices, it is better to use stereo input than rgbd input. There may be improvements in the future, such as https://github.com/luxonis/depthai-core/issues/875 or see if [DisparityWLSFilter](https://docs.opencv.org/4.5.0/d9/d51/classcv_1_1ximgproc_1_1DisparityWLSFilter.html) helps.
A little benefit of disabling Subpixel is that 12 shave cores can be used for NeuralNetwork. So I compiled [a new SuperPoint blob for 6 shave cores](https://github.com/borongyuan/depthai-superpoint/tree/main/blobs).

matlabbe:
Thanks for the MR!

PR #1116, opened 2023-08-19
Update OAK stereo config and IMU rate

borongyuan:
The kernel size of census transform is set to AUTO by default. Looks like CT-7x7 is used at 800p. I found that CT-7x9 gives better depth quality for both 800p and 400p. Don't know why it is not used as default.
IMU rate is set to 100 Hz as none of the devices with BMI270 can reach 200 Hz. PoE models can go up to around 120 Hz. It works well when using msckf_vio. Feature noise is best reduced to 0.02 to avoid drift.
Msckf_vio may not be the most accurate VIO, but it is certainly robust. I often encounter drift issues when using OpenVINS, and I am still looking for the cause.


matlabbe:
Thanks for all this information, very useful!

PR #1137, opened 2023-09-19
Add Histogram Equalization

borongyuan:
Both OpenVINS and ORB-SLAM3 have histogram equalization methods. ORB-SLAM3 says that it is mainly for TUM-VI dataset. When I used the OAK camera for outdoor testing, I found that the contrast of the original images is actually low. The SLAM performance is improved after applying histogram equalization.


matlabbe:
Great, thx!

PR #1142, opened 2023-10-05
Update odometry info of OpenVINS

borongyuan:
Add openvins msckf features as outliers for visualization. However, msckf features are not entirely outliers. They are actually lost features + long-term features not used in updaterSLAM + short-term features. They are used in updaterMSCKF.
https://github.com/rpng/open_vins/blob/4534a2f32d4763bdc8c95121b3292c7423e12aca/ov_msckf/src/core/VioManager.cpp#L497-L500

Fill localMap with active tracked features. Active tracked features are more numerous and have corresponding IDs. We can see how they are triangulated.


matlabbe:
Thx!

PR #1143, opened 2023-10-07
OAK bug fixes and improvements

borongyuan:
Fix distortion correction problem of small FOV devices.
I have been testing with wide FOV devices before. When I tried OAK-D S2 today I found that it didn't work at all. I checked [the doc](https://docs.luxonis.com/projects/api/en/latest/references/cpp/#_CPPv4N3dai21StereoDepthProperties26useHomographyRectificationE) and found that the default configuration does not perform distortion correction on small FOV devices. I can't understand why they do this. We need to force distortion correction on all devices.

Remove deprecated IMU firmware update API.
enableFirmwareUpdate() is deprecated since https://github.com/luxonis/depthai-core/commit/12b198a67f9c7a56747666679f00763c35cf64e3. I don't feel the need to deal with firmware updates here. Most devices don't need to be updated. Or the user should have updated it in advance.

Add useSpecTranslation option.
We have had some discussions about this parameter before (https://github.com/luxonis/depthai-core/issues/358 & https://github.com/luxonis/depthai-core/issues/847). But most users are probably unaware of this potential problem. Although they insist that the translation information from the board design data will have better accuracy in most cases. But I actually have a development version of OAK that is not for sale. They changed the color camera's FOV, global shutter, and IMU model according to my requirements. The camera modules are all glued on by hand, so I don't think there's any manufacturing precision. I also noticed that the RAE's lens looks a bit skewed as well, although I haven't tested it yet. So let’s use the calibration data by default. It guarantees at least a lower bound of accuracy and meets most users' expectations.
From my experience, this parameter may have an impact on the accuracy and robustness of stereo VIO. If you read [OpenVINS paper](https://pgeneva.com/downloads/papers/Geneva2020ICRA.pdf), you will find that mono mode often performs better than stereo mode. The accuracy of the baseline and different feature representations have a certain impact.

Reduce depth image noise and simplify calculation for compressed transport.
Depth images have fewer textures than mono images and therefore have higher compression rates when using mjpeg compression. So we can set the quality to the highest level without consuming too much bandwidth. The advantage of this is that it can reduce the noise generated near the sharp edges during depth map compression. Especially at the edges where depth values are missing. These noises are mainly caused by DCT during jpeg encoding. This doesn't completely eliminate the noise, but it's better than before. It should be better to use lossless jpeg mode, but currently OpenCV is unable to decode it. We will see if there are other decoding methods later, such as using ffmpeg or nvjpeg.
[cv::divide()](https://docs.opencv.org/4.x/d2/de8/group__core__array.html#ga6db555d30115642fedae0cda05604874) is supposed to be more efficient than the forEach method. It can handle division by 0.

### DIFF
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -56,8 +56,8 @@ public:
 	virtual ~CameraDepthAI();
 
 	void setOutputDepth(bool enabled, int confidence = 200);
+	void setUseSpecTranslation(bool useSpecTranslation);
 	void setAlphaScaling(float alphaScaling = 0.0f);
-	void setIMUFirmwareUpdate(bool enabled);
 	void setIMUPublished(bool published);
 	void publishInterIMU(bool enabled);
 	void setLaserDotBrightness(float dotProjectormA = 0.0f);
@@ -83,8 +83,8 @@ private:
 	bool outputDepth_;
 	int depthConfidence_;
 	int resolution_;
+	bool useSpecTranslation_;
 	float alphaScaling_;
-	bool imuFirmwareUpdate_;
 	bool imuPublished_;
 	bool publishInterIMU_;
 	float dotProjectormA_;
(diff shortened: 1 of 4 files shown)


matlabbe:
My RAE has been shipped, next weekend I will be able to test more all your updates. Thx a lot keeping depthai integration up to date!

PR #1145, opened 2023-10-11
Supplement and Adjust OpenVINS Parameters

borongyuan:
Hi, @matlabbe
I added more OpenVINS params. Enabling online calibration sometimes improves accuracy, but may also hurt the performance due to degenerate motions. You can refer to the conclusion of [this video](https://www.youtube.com/watch?v=uf367hP6s3M). Dynamic initialization may be useful for odom reset. Even if it is enabled, the system will try static initialization first. There are also several feature initializer params which may need to be tuned in outdoor scenes.
There’s no need to merge this PR right away. There is another important issue to resolve. We are not using covariance properly. When I was testing in some large scenes recently, I found that after successful loop closure, some neighbor links would be stretched extremely long. And VIO has been working correctly, so I thought it should be a covariance problem. I checked other issues and found that we should use twist covariance, but now pose covariance is set. We should not be able to use get_marginal_covariance() to obtain twist covariance. Because v_IinI and w_IinI are not in the state vector, there is only v_IinG. It seems that twist covariance can only be obtained using fast_state_propagate(). Then we need to rotate it into the base frame. Do I understand correctly?

### DIFF
--- a/corelib/include/rtabmap/core/Parameters.h
+++ b/corelib/include/rtabmap/core/Parameters.h
@@ -581,7 +581,11 @@ class RTABMAP_CORE_EXPORT Parameters
     RTABMAP_PARAM(OdomOpenVINS, UseKLT,                    bool,   true,   "If true we will use KLT, otherwise use a ORB descriptor + robust matching");
     RTABMAP_PARAM(OdomOpenVINS, NumPts,                    int,    200,    "Number of points (per camera) we will extract and try to track");
     RTABMAP_PARAM(OdomOpenVINS, MinPxDist,                 int,    15,     "Eistance between features (features near each other provide less information)");
-    RTABMAP_PARAM(OdomOpenVINS, KNNRatio,                  double, 0.7,    "Descriptor knn threshold for the top two descriptor matches");
+    RTABMAP_PARAM(OdomOpenVINS, FiTriangulate1d,           bool,   false,  "If we should perform 1d triangulation instead of 3d");
+    RTABMAP_PARAM(OdomOpenVINS, FiRefineFeatures,          bool,   true,   "If we should perform Levenberg-Marquardt refinement");
+    RTABMAP_PARAM(OdomOpenVINS, FiMaxRuns,                 int,    5,      "Max runs for Levenberg-Marquardt");
+    RTABMAP_PARAM(OdomOpenVINS, FiMaxBaseline,             double, 40,     "Max baseline ratio to accept triangulated features");
+    RTABMAP_PARAM(OdomOpenVINS, FiMaxCondNumber,           double, 10000,  "Max condition number of linear triangulation matrix accept triangulated features");
 
     RTABMAP_PARAM(OdomOpenVINS, UseFEJ,                    bool,   true,   "If first-estimate Jacobians should be used (enable for good consistency)");
     RTABMAP_PARAM(OdomOpenVINS, Integration,               int,    1,      "0=discrete, 1=rk4, 2=analytical (if rk4 or analytical used then analytical covariance propagation is used)");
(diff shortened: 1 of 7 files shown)


corelib/include/rtabmap/core/Parameters.h
```diff
@@ -628,56 +646,56 @@ class RTABMAP_CORE_EXPORT Parameters
...
-    RTABMAP_PARAM(Vis, CorFlowWinSize,           int, 16,     uFormat("[%s=1] See cv::calcOpticalFlowPyrLK(). Used for optical flow approach.", kVisCorType().c_str()));
-    RTABMAP_PARAM(Vis, CorFlowIterations,        int, 30,     uFormat("[%s=1] See cv::calcOpticalFlowPyrLK(). Used for optical flow approach.", kVisCorType().c_str()));
-    RTABMAP_PARAM(Vis, CorFlowEps,               float, 0.01, uFormat("[%s=1] See cv::calcOpticalFlowPyrLK(). Used for optical flow approach.", kVisCorType().c_str()));
-    RTABMAP_PARAM(Vis, CorFlowMaxLevel,          int, 3,      uFormat("[%s=1] See cv::calcOpticalFlowPyrLK(). Used for optical flow approach.", kVisCorType().c_str()));
+    RTABMAP_PARAM(Vis, MaxFeatures,               int,   1000,  "0 no limits.");
+    RTABMAP_PARAM(Vis, MaxDepth,                  float, 60.0,  "Max depth of the features (0 means no limit).");
+    RTABMAP_PARAM(Vis, MinDepth,                  float, 0.1,   "Min depth of the features (0 means no limit).");
```
matlabbe: A particular reason why they need to be changed? I would prefer keeping the whole defaults. In stereo mode, using features very far (or even without depth) can be useful for better orientation estimation. For the min depth, is it really a problem that the limit is 0? (would not limit VSLAM application using very small baseline, tracking very close objects)

borongyuan:
Although OpenVINS will accept features within a certain range in FeatureInitializer. But it only knows the distance of the features when they are triangulated. In the feature detection and tracking stage, if we want to exclude some image areas, such as parts occluded by the robot or vehicle body, we need to use masks. Now we have two mask methods. One is to use pre-drawn mask images like original OpenVINS. The other is to borrow the kVisDepthAsMask parameter in RTAB-Map. This makes the configuration more convenient and the range is adjustable. Using masks can reduce the outlier ratio of feature detection and tracking, making RANSAC more likely to be correct.

borongyuan:
https://github.com/introlab/rtabmap/blob/f06cc0b17f49209c19107663b45fed54433e8f7f/corelib/src/odometry/OdometryMSCKF.cpp#L941
What is this issue? Has it been solved yet?

matlabbe:
> https://github.com/introlab/rtabmap/blob/f06cc0b17f49209c19107663b45fed54433e8f7f/corelib/src/odometry/OdometryMSCKF.cpp#L941
> 
> 
> What is this issue? Has it been solved yet?

Good question, not sure if it was a bug in g2o or in the integration with g2o. We would need to try with latest g2o version to see if there is a problem. Note that I didn't use that code for the past 5 years (and never tested live on a camera, just on euroc dataset)

corelib/include/rtabmap/core/Parameters.h
borongyuan: Too many params were introduced when improving OpenVINS integration. So I try to reuse existing params whenever possible, as long as they have almost the same meaning. I modified these default values previously to make them compatible with OpenVINS params. But I now found a better implementation and changed them back.

0.1-60.0 is the default used by OpenVINS. This is basically beyond the theoretical measurement range of most sensors. So I previously thought that this should not affect the performance in various configurations. OpenVINS considers points that are too close as bad features, probably because of focus reasons. For a fixed-focus lens, if it focuses at infinity, close objects will be more or less out of focus.

borongyuan:
I've fixed the covariance part. Now these can be merged.

corelib/include/rtabmap/core/Parameters.h
matlabbe: Understood!

matlabbe:
Great thank you!

PR #1150, opened 2023-10-24
Add missing filterKeypointsByDepth step

borongyuan:
filterKeypointsByDepth is missing in this if-branch

### DIFF
--- a/corelib/src/Features2d.cpp
+++ b/corelib/src/Features2d.cpp
@@ -732,19 +732,19 @@ std::vector<cv::KeyPoint> Feature2D::generateKeypoints(const cv::Mat & image, co
 		for (int j = 0; j<gridCols_; ++j)
 		{
 			cv::Rect roi(globalRoi.x + j*colSize, globalRoi.y + i*rowSize, colSize, rowSize);
-			std::vector<cv::KeyPoint> sub_keypoints;
-			sub_keypoints = this->generateKeypointsImpl(image, roi, mask);
-			limitKeypoints(sub_keypoints, maxFeatures);
+			std::vector<cv::KeyPoint> subKeypoints;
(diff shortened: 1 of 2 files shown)


borongyuan:
Keypoints with no valid 3D points are mostly because they are too far away to be measured. filterKeypointsByDepth() will remove them. But this doesn't seem to be what we want. Should we consider retaining this part?

matlabbe:
The features should be already filtered by depth there if `_depthAsMask` is true and if RGB-D data is provided. For stereo, that could be indeed added. Suggestion:
```cpp
if(depthMask.empty() && (_feature2D->getMinDepth() > 0.0f || _feature2D->getMaxDepth() > 0.0f))
```

borongyuan:
> ```c++
> if(depthMask.empty() && (_feature2D->getMinDepth() > 0.0f || _feature2D->getMaxDepth() > 0.0f))
> ```

I noticed this doesn't work with ORBOctree.
https://github.com/introlab/rtabmap/blob/71bb0cf2262ed73c937312da3b3c7cc8d0ad0f13/corelib/src/opencv/ORBextractor.h#L66

Do other types of features have similar problems?

matlabbe:
> > ```c++
> > if(depthMask.empty() && (_feature2D->getMinDepth() > 0.0f || _feature2D->getMaxDepth() > 0.0f))
> > ```
> 
> I noticed this doesn't work with ORBOctree.
> 
> https://github.com/introlab/rtabmap/blob/71bb0cf2262ed73c937312da3b3c7cc8d0ad0f13/corelib/src/opencv/ORBextractor.h#L66
> 
> Do other types of features have similar problems?

I added a new commit to fix OrbOctree ignoring the mask. I think it was the only one ignoring it. I added my suggestion. It seems to work as expected now.

PR #1154, opened 2023-11-02
transform features after source images decimation

borongyuan:


matlabbe:
Indeed, for depthai features! Thx

PR #1159, opened 2023-11-14
fix protocol detection for undiscoverable devices

borongyuan:
In some network topologies PoE cameras cannot be discovered using getAllAvailableDevices(). But if their IPs are manually specified, they are actually accessible. This fix allows us to use these cameras. But there are still some small problems. For example, we cannot guarantee that the PoE camera is actually found before dai::Device is initialized. This is because there is only getDeviceByMxId() in the API, but not getDeviceByName(). Before dai::Device is initialized, its protocol cannot be detected.


matlabbe:
Great, thx!

PR #1173, opened 2023-11-28
improve NMS implementation

borongyuan:
This fixes the problem of extracting multiple keypoints at the same location. Previously, where prob is greater than the threshold were extracted as keypoints.
https://github.com/introlab/rtabmap/blob/7ab1ee0a0b12dd8d5f607c5cf536505b5d0d7779/corelib/src/camera/CameraDepthAI.cpp#L653
https://github.com/introlab/rtabmap/blob/7ab1ee0a0b12dd8d5f607c5cf536505b5d0d7779/corelib/src/superpoint_torch/SuperPoint.cc#L161
Therefore in NMS, local maximums are found first, and then neighbors are suppressed. Now for loop closure, it works fine when the image resolution is 400P or 800P. For odometry, 800P is not suitable for F2M, but is suitable for F2F. Because the position of keypoints is more jittery.

### DIFF
--- a/corelib/src/util2d.cpp
+++ b/corelib/src/util2d.cpp
@@ -2118,6 +2118,7 @@ void NMS(
     cv::Mat inds = cv::Mat(cv::Size(img_width, img_height), CV_16UC1);
 
     cv::Mat confidence = cv::Mat(cv::Size(img_width, img_height), CV_32FC1);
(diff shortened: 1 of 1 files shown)


matlabbe:
Alright, thx!

PR #1192, opened 2024-01-01
Reimplement NMS using morphological operations

borongyuan:
This borrows from the PyTorch implementation in SuperGlue, but uses this [corrected version](https://github.com/magicleap/SuperGluePretrainedNetwork/issues/112#issuecomment-1199862202). This implementation is relatively simple, and since OpenCV has various optimizations for morphological operations, it will theoretically have better performance. But I haven't done any performance comparison testing yet. Since this is performed on heatmap, the NMS in SuperPoint Torch can be changed to PyTorch implementation. This way we can remove this part from util2d and introduce ANMS later.

### DIFF
--- a/corelib/src/camera/CameraDepthAI.cpp
+++ b/corelib/src/camera/CameraDepthAI.cpp
@@ -679,25 +679,39 @@ SensorData CameraDepthAI::captureImage(CameraInfo * info)
 		auto heatmap = features->getLayerFp16("heatmap");
 		auto desc = features->getLayerFp16("desc");
 
-		cv::Mat prob(200, 320, CV_32FC1, heatmap.data());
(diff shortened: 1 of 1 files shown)


matlabbe:
The NMS version for SuperPoint Torch you are referring is this one? https://github.com/introlab/rtabmap/blob/d135ef02e2fdce19ce8484f3ad45428042e1e773/corelib/src/superpoint_torch/SuperPoint.cc#L168-L192

I am not using that much that code right now, but if you use it and know how to update it, you can make another PR.

borongyuan:
> I am not using that much that code right now, but if you use it and know how to update it, you can make another PR.

I don't use this part very much either. But I will test the new environment after upgrading my Jetson system to JetPack 6 in the near future and update this part.

PR #1193, opened 2024-01-03
OAK Internal Sync & DepthAI HF-Net Support

borongyuan:
All messages except IMU are synchronized using the Sync node introduced in [v2.24.0](https://github.com/luxonis/depthai-core/releases/tag/v2.24.0). LGTM, but I'll keep this PR as a draft until the depthai in ROS repo is updated to 2.24.0. In addition, they have not added doc for Sync node yet, only providing new examples. We can do more inspections and tests during this time.
The IR settings have also been changed to new methods in v2.24.0.

### DIFF
--- a/corelib/include/rtabmap/core/camera/CameraDepthAI.h
+++ b/corelib/include/rtabmap/core/camera/CameraDepthAI.h
@@ -59,7 +59,7 @@ public:
 	void setDepthProfile(int confThreshold = 200, int lrcThreshold = 5);
 	void setRectification(bool useSpecTranslation, float alphaScaling = 0.0f);
 	void setIMU(bool imuPublished, bool publishInterIMU);
-	void setIrBrightness(float dotProjectormA = 0.0f, float floodLightmA = 200.0f);
+	void setIrIntensity(float dotIntensity = 0.0f, float floodIntensity = 0.0f);
 	void setDetectFeatures(int detectFeatures = 0);
 	void setBlobPath(const std::string & blobPath);
 	void setGFTTDetector(bool useHarrisDetector = false, float minDistance = 7.0f, int numTargetFeatures = 1000);
@@ -86,8 +86,8 @@ private:
 	float alphaScaling_;
 	bool imuPublished_;
 	bool publishInterIMU_;
-	float dotProjectormA_;
-	float floodLightmA_;
+	float dotIntensity_;
+	float floodIntensity_;
 	int detectFeatures_;
 	bool useHarrisDetector_;
 	float minDistance_;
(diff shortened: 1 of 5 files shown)


borongyuan:
Since I have recently received several new OAK cameras equipped with BNO086, I presume they have stopped producing the ones equipped with BMI270. According to the document and datasheet, the accelerometer and gyroscope of the BNO086 cannot be set to the same sample rate.
> - Accelerometer: 15Hz, 31Hz, 62Hz, 125Hz, 250Hz 500Hz
> - Gyroscope: 25Hz, 33Hz, 50Hz, 100Hz, 200Hz, 400Hz
> - Magnetometer: 100Hz

I tested different configs. Either
```
imu->enableIMUSensor({dai::IMUSensor::ACCELEROMETER_RAW, dai::IMUSensor::GYROSCOPE_RAW}, 100);
```
or
```
imu->enableIMUSensor(dai::IMUSensor::ACCELEROMETER_RAW, 125);
imu->enableIMUSensor(dai::IMUSensor::GYROSCOPE_RAW, 100);
```
will cause the IMU data to be output at 100Hz. Part of the accelerometer data will be discarded directly to reduce to 100 Hz without linear interpolation. In this configuration, the VIO's robustness is affected and may begin to drift during sudden stops. This problem can be greatly avoided by setting the IMU output rate to 200 Hz. This can be compatible with both BMI270 and BNO086 devices. But it may not be the optimal configuration for both types of devices. BMI270 is actually unable to stably reach an output of 200 Hz, generally only 160Hz~180Hz. If we only consider supporting BNO086, we can actually
```
imu->enableIMUSensor({dai::IMUSensor::ACCELEROMETER_RAW, dai::IMUSensor::GYROSCOPE_RAW, dai::IMUSensor::ROTATION_VECTOR}, 200);
```
In this way, we directly use the orientation output of BNO086 that fuses 9-axis data, without using Madgwick filter or complementary filter later.

borongyuan:
Set sync threshold to half of the frame interval. This avoids incorrect sync between different cameras.

borongyuan:
The depthai-core in the ROS repository is still at 2.23.0. I will remind them.

matlabbe:
Ah yeah I didn't verify if the ros package was released with 0.24 (just assumed it since it is almost 3 months ago old), I created a label to quickly know that we are waiting for dependency update.

matlabbe:
depthai 2.24 binaries are out: 
* http://repositories.ros.org/status_page/noetic_default.html?q=depthai
* http://repo.ros2.org/status_page/ros_humble_default.html?q=depthai

PR #1203, opened 2024-01-21
support color histogram equalization

borongyuan:
Add support for histogram equalization of color images. Color images are first converted to YCrCb color space, and histogram equalization is performed only on the Y channel.
Why use YCrCb instead of HSV or LAB?
They should have similar results. But according to [OpenCV's color conversion documentation](https://docs.opencv.org/4.x/de/d25/imgproc_color_conversions.html), the Y channel of BGR2YCrCb is calculated in the same way as BGR2GRAY. Therefore, it can ensure the same effect as converting to gray image first and then performing histogram equalization.


matlabbe:
LGTM!

PR #1213, opened 2024-02-04
SuperPoint Torch Simple NMS

borongyuan:
Successor of #1192 

I did't actually seen SuperPoint Torch perform very well on the Jetson Orin Nano if it's simply called through libtorch. After all, its FP32 computing power is not very good, only 1.28 TFLOPS. We may consider adding some optimizations later.


matlabbe:
LGTM!

PR #1223, opened 2024-02-26
not limitKps inside Feature2D::generateKeypoints when using pyDetector

hellovuong:
This only happens with SuperPoint. Rtabmap will limit key points if exceeds the limit set by Kp/MaxFeatures THEN it will generate descriptors based on extracted key points. It will work classic Features Extraction method. However, SuperPoint will generate key points and descriptors at the same time, if the limit key points method calls inside the extract key point method it will lead to a mismatched number of key points and descriptors later when the generate descriptor method calls. It is solved by ignoring the limit key points method inside the generate key point method and only limiting key points after generating the descriptors if necessary.

### DIFF
--- a/corelib/src/Features2d.cpp
+++ b/corelib/src/Features2d.cpp
@@ -741,13 +741,13 @@ std::vector<cv::KeyPoint> Feature2D::generateKeypoints(const cv::Mat & image, co
 			if(roi.x || roi.y)
 			{
 				// Adjust keypoint position to raw image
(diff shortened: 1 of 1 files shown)


hellovuong:
Sorry, there is a wrong pull. I will fix it soon!

matlabbe:
Good catch. I'll merge it now as it can cause errors, but I opened a new issue https://github.com/introlab/rtabmap/issues/1225 to think about supporting that feature for PyDetector. To use a limit, a workaround would be to limit the keypoints/descriptors inside the python file and set `Vis/MaxFeatures=0`. For the visual word dictionary, `Kp/MaxFeatures` should still work correctly with this patch.

Issue #1215, opened 2024-02-06
[Feature-Request] HF-Net


borongyuan:
@matlabbe @cdb0y511 
Hi, when I further looked at the work related to SuperPoint, I noticed that it seemed that [HF-Net](https://projects.asl.ethz.ch/datasets/doku.php?id=cvpr2019hfnet) could also be added. It provides global descriptors for coarse localization in addition to SuperPoint.
![](https://raw.githubusercontent.com/ethz-asl/hfnet/master/doc/assets/hfnet.jpg)

RTAB-Map reserves an interface for global descriptor, but does not implement related functions  (#1105). Therefore, I am thinking about whether this part can be implemented universally and how to combine it with the existing memory management mechanism of RTAB-Map. My initial feeling is that it can help retrieval, and secondly it may be further used to improve loop closure hypotheses.

The difficulty in integrating HF-Net is that it is implemented in TensorFlow 1. I only have a few old devices with the required environment. So I also converted it to ONNX first. But I'm more curious about whether it will work on OAK. That would be really fun if it could. This model file looks a bit large, I guess it is mainly because of the fully connected layer at the end of the global head. Even if OAK can't run it, I will try to prune it into a MobileNet based SuperPoint implementation. Some people have tried to train SuperPoint based on MobileNet directly but did not achieve good results. If OAK can run it, I will not develop the inference solution on the host for the time being. But if you are interested, you can also develop it. This is an ONNX file that I converted but haven't tested yet.
https://drive.google.com/file/d/1xKJi1RfKZTLQ0JACD4H76HRpxr6kFgtn/view?usp=sharing
![hfnet_480x640 onnx](https://github.com/introlab/rtabmap/assets/22918715/038b7732-5852-47a0-825f-c2ff0220fbdd)

hellovuong:
Hi, I am also looking at implementing of global descriptor for rtabmap. There are some repos already implemented in HF-NET TensorRT also, I tested it. Should we consider that? should not be so hard to add to the rtabmap codebase. I just don't know how to integrate it with RTABMap memory management for image retrieval.

matlabbe:
[...]

Issue #1105, opened 2023-08-03
about GlobalDescriptor 


frischzenger:
GlobalDescriptor was not implemented.

frischzenger:
the image GlobalDescriptor was not implemented.

matlabbe:
Is there a specific question? The GlobalDescriptor is currently a placeholder, there is no built-in mechanism to create one or even use it to find loop closures. It was introduced to help integration with https://github.com/MarvinStuede/cmr_lidarloop, so that GlobalDescriptor can be saved in the database, and then re-extracted afterwards for external loop closure detection.

borongyuan:
I recently re-read previous papers to consider how to utilize the global descriptor, especially VLAD. It‘s mainly useful in two aspects, evaluating the similarity between nodes, and fast nodes retrieval.

Currently, the similarity is evaluated by
https://github.com/introlab/rtabmap/blob/a0476af896f43361124be875925114305766b415/corelib/src/Signature.cpp#L234
Likelihood is calculated via similarity or TF-IDF. With global descriptors of two nodes, similarity can be obtained by directly calculating their scalar product. Especially when VLAD has been normalized, the result is actually cosine similarity.

Because VLAD can encode images into very small data, global descriptors for a large number of nodes can be kept in memory for fast retrieval. For the 4096-D VLAD output by NetVLAD or HF-Net, each global descriptor will occupy 16KB. So if there are 10,000 nodes, it would take 164MB to store their global descriptors. KNN search can be used to find nodes that may detect loop closure from LTM, or to filter nodes used to calculate likelihood from WM, similar to LoopGPS.

Another question is whether Bayesian filter is still necessary when using global descriptor. I currently believe that evaluating likelihood is more reasonable than using similarity directly. After all, it is a bit difficult to determine thresholds for different types of global descriptors. Likelihood can tell whether the similarity is truly significant, and eliminate unnecessary comparisons.

We can now keep global descriptors all in memory. However, the memory management of RTAB-Map and VLAD make it possible to challenge extremely large-scale scenarios (such as city-level) in the future. By then we may also consider memory management of global descriptors.

matlabbe:
The Bayes filter is still useful to filter spurious high likelihood (false positives), i.e., we need some consecutive high likelihood in an area to trigger a loop closures/re-localization.

[...]

PR #1263, opened 2024-04-13
copy global descriptor

borongyuan:
The inference is correct using HF-Net blob file. I made a silly mistake when converting to cv::Mat. This solves https://github.com/introlab/rtabmap/issues/1215#issuecomment-2002006330. Now I can get similar results to https://github.com/introlab/rtabmap/issues/1105#issuecomment-2031099209 in online testing.


matlabbe:
Great, thx!

PR #1272, opened 2024-04-28
adjust likelihood using z-score

borongyuan:
I constructed a new method to adjust the likelihood, as a solution to https://github.com/introlab/rtabmap/issues/1105#issuecomment-2031099209. It should be compatible with previous similarity evaluation methods, as well as VLAD. The idea is to calculate z-score for values ​​greater than μ + σ. This eliminates the effect of different distributions. For new location likelihood, how significant the max value is will be evaluated.

### DIFF
--- a/corelib/src/Rtabmap.cpp
+++ b/corelib/src/Rtabmap.cpp
@@ -5265,15 +5265,14 @@ void Rtabmap::adjustLikelihood(std::map<int, float> & likelihood) const
 
 
 	//Adjust likelihood with mean and standard deviation (see Angeli phd)
-	float epsilon = 0.0001;
 	float max = 0.0f;
 	int maxId = 0;
 	for(std::map<int, float>::iterator iter=++likelihood.begin(); iter!= likelihood.end(); ++iter)
 	{
 		float value = iter->second;
-		if(value > mean+stdDev && mean)
+		if(value > mean+stdDev && stdDev)
 		{
-			iter->second = (value-(stdDev-epsilon))/mean;
+			iter->second = (value-mean)/stdDev;
 			if(value > max)
 			{
 				max = value;
@@ -5295,9 +5294,9 @@ void Rtabmap::adjustLikelihood(std::map<int, float> & likelihood) const
 		}
 	}
 
-	if(stdDev > epsilon && max)
+	if(max > mean)
 	{
-		likelihood.begin()->second = mean/stdDev + 1.0f;
+		likelihood.begin()->second = stdDev/(max-mean) + 1.0f;
 	}
 	else
 	{


borongyuan:
Yes, this change needs more testing. I've only done some preliminary testing so far, and found it produces pretty pleasing likelihood curves. In particular, no matter which method is used to calculate the raw likelihood, the likelihood curves are very similar. It’s just that the PDF curves have become a bit flat, so we may need to further adjust the parameters of the Bayesian filter.

The reason is that z-score is not prone to particularly large values. A significant loop event generally only has 3 to 6 σ. New location likelihood is limited to between 1 and 2. Since such distribution characteristics are stable, Bayesian filtering should be able to handle them well.

matlabbe:
I added a new parameter to select this approach: set `Rtabmap/VirtualPlaceLikelihoodRatio=1` to use approach of this MR. 

Here is a comparison of resulting hypotheses of the Bayes filter based on the old (original) and proposed approaches. Note that for the proposed approach, we should set `Rtabmap/LoopThr` to `0.5` to be somewhat equivalent to `0.11` (default) using the original approach. On right is the Precision/Recall curve (Recall on x-axis, Precision on y-axis).

NewCollege dataset:
 * `Rtabmap/VirtualPlaceLikelihoodRatio=0` (original approach)
    NewCollegeOld
 * `Rtabmap/VirtualPlaceLikelihoodRatio=1` (proposed approach)
    NewCollege

CityCentre dataset:
 * `Rtabmap/VirtualPlaceLikelihoodRatio=0` (original approach)
    CityCentreOld
 * `Rtabmap/VirtualPlaceLikelihoodRatio=1` (proposed approach)
    CityCentre

UdeS_1Hz dataset:
 * `Rtabmap/VirtualPlaceLikelihoodRatio=0` (original approach)
    UdeS_1HzOld
 * `Rtabmap/VirtualPlaceLikelihoodRatio=1` (proposed approach)
    UdeS_1Hz

In conclusion, the results are similar using one or the other approach for BOW. However, the `Rtabmap/LoopThr` seems to vary more between the datasets, which is why I didn't put the proposed approach to be new default. I'll try to do the same comparison in coming weeks with global descriptor NETVLAD with which I expect to get better Precision/Recall with `Rtabmap/VirtualPlaceLikelihoodRatio=1`.

borongyuan:
Great to see your test results. I've been doing real-time testing with the OAK camera recently, so I haven't tested it on these datasets yet. I'm a little surprised that from your PR curves, it seems that the AUC is higher even if you still use BOW?
I'm still having some difficulty getting more significant posterior from Bayesian filtering. Intuitively, it can be clearly seen that a longer path keeping high likelihood is needed to filter to a significant result. NETVLAD does have better recall capabilities than BOW at some locations, but the current parameters don't seem to be unleashing its potential.

PR #1294, opened 2024-05-30
Add SeerSense Support

borongyuan:
Hi @matlabbe,

XVisio invited me to try out their [SeerSense](https://www.xvisiotech.com/product/seersense-ds80/) depth camera last year. Their product is similar to OAK in many aspects. For example, its processor uses Myriad X, stereo cameras has 150° DFOV, and it supports OpenVINO. Besides, it has some highlights, such as having 2 deep engines (SGBM/ToF) and providing edge VIO (which OAK has not implemented yet). I have now basically integrated it into RTAB-Map (including RGBD and Odom), and the performance looks good.

[Here](https://www.xvisiotech.com/xvisio-sdk-doc/06_xvisio_api/06_00_xvisio_api/index.html) is the documentation for their C++ API. I think their architecture design is simpler, but not as deeply configurable as DepthAI. This may limit some of its uses.

I sent you an invitation email so that they can send you a module for test and review.


borongyuan:
https://github.com/introlab/rtabmap/assets/22918715/ba15f863-8d50-44d2-bba4-b08edd91d15d

matlabbe:
That looks great! I do have a xvisio camera here (that they sent me in 2019), but I think it is more a prototype and I had some issues with the xslam sdk (old name of their sdk). I don't think I can use it with the new SDK.

PR #1299, opened 2024-06-14
Add flags for OpenNI

Serafadam:
Adds flag WITH_OPENNI to mimic how other interfaces are enabled/disabled. Not having it actually caused crashes similar to https://github.com/CognitiveRobotics/organized_segmentation_tools/issues/1 https://github.com/introlab/rtabmap_ros/issues/117 when building on older systems.


matlabbe:
Great thx!

PR #1328, opened 2024-08-18
Ouster SDK integration

matlabbe:


matlabbe:
On Windows, use OusterSDK DLL version: https://github.com/ouster-lidar/ouster-sdk/pull/608 to avoid "silent crash" (in this [function](https://github.com/ouster-lidar/ouster-sdk/blob/27bc78f5aeccb3b74e8860ee7fad57f0e128aa49/ouster_osf/include/ouster/osf/metadata.h#L420-L423)) with the static library (still not sure what is causing this, rtabmap includes a bunch of dependencies that may clash with the static library).

For convenience, apply this patch:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9994531..4986fc6 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ include(DefaultBuildType)
 include(VcpkgEnv)

 # ==== Project Name ====
-project(ouster_example VERSION 20231031)
+project(OusterSDK VERSION 20231031)
```
so that OusterSDK library installed in `Program Files` can be detected automatically with cmake, otherwise we need to do:
```
cmake -DOusterSDK_DIR="C:/Program Files (x86)/ouster_example/lib/cmake/OusterSDK ..
```
Well, if `C:\Program Files (x86)\OusterSDK\bin` is on the `PATH`, that may work even if we don't change project name (not tested).

Issue #1320, opened 2024-08-02
Add very low RAM mode for iOS App large-scale scanning


matlabbe:
Like android app's "Trajectory Mode", we could do something similar for iOS. This could save battery and make it able to scan in order of a hour or more. See this thread for more info: http://official-rtab-map-forum.206.s1.nabble.com/iOS-best-practices-for-only-tracking-long-recordings-td10300.html

zeeyang:
Is it possible to post process the recorded data on device to generate optimized poses? And what's the equivalent in CLI? Is it `rtab-reprocess`?

matlabbe:
I didn't implement a way to reprocess on device. 

The GUI way to reprocess is to open RTAB-Map Desktop, open Preferences->Source, then select Database as input source. Scroll down to select the path of the recorded database. Say yes to use odometry from the database, but no for setting rtabmap detection rate to 0. Check database stamps for framerate. Go back on main UI and press play.

To use the CLI tool:
```
rtabmap-reprocess \
   --Mem/RehearsalSimilarity 0.3 \
   --Kp/MaxFeatures 400 \
   --Rtabmap/MaxRetrieved 2\
   --RGBD/MaxLocalRetrieved  2\
   --Mem/MapLabelsAdded true \
   --Rtabmap/MemoryThr 0 \
   --Mem/STMSize 10 \
   --RGBD/ProximityBySpace true \
   --RGBD/LinearUpdate 0.05 \
   --RGBD/AngularUpdate 0.05 \
    recording.db output.db
```

matlabbe:
Closing as already available on iOS app under Menu -> Advanced -> New Data Recording.

Issue #1281, opened 2024-05-15
Rtabmap on Android devices


Genozen:
Hello, I'm comparing the ios and android version of the Rtabmap and noticed something interesting on the android phones:
![image](https://github.com/introlab/rtabmap/assets/74474611/4dfde3ad-f08b-4cbb-956b-fc18ec7c0ed5)

Sometimes the pose jitters a lot. I've searched around the issues and forums but have not seen related mention about this.

Any idea as to why this happens?

matlabbe:
Which version are you using? It looks like the pose is jumping between two referentials, there could be a bug. We do offset the origin of the received pose, maybe there is a offset missing sometimes. Let me know which [APK](https://github.com/introlab/rtabmap/wiki/Installation#rtab-map-arcore-apk) you are using.

Genozen:
@matlabbe  I used the tutorial from here is: https://github.com/introlab/rtabmap/wiki/Installation#rtab-map-arcore-apk.

matlabbe:
Reproduced with ARCore-Java driver by moving fast:
<img src="https://github.com/introlab/rtabmap/assets/2319645/f69cfcfb-5056-4672-8b92-719913610b6e" width="200"/>

If we move slow it is fine:
<img src="https://github.com/introlab/rtabmap/assets/2319645/eacec785-74b0-4945-bbb0-69592b9b5c18" width="200"/>

There could be a bug on how arcore localizations are filtered, causing the back-and-forth jumps. In [commit](https://github.com/introlab/rtabmap/commit/c5c190a2a06ca7b1f77cbe45ae6c31f83cb90067) above, I disabled that filtering feature by default. On your version without reinstalling, you can change this setting under Mapping section below:

<img src="https://github.com/introlab/rtabmap/assets/2319645/4da55b04-9d69-4bfd-a272-92a1e6aa27bf" width="200"/>

PR #1340, opened 2024-09-09
iOS various updates 

matlabbe:
- [x] Added Data Recording Mode. https://github.com/introlab/rtabmap/issues/1320
- [x] Added option to filter ARKit localization jumps.
  *  TODO: find a default threshold working for most usages (maybe instead thresholding on speed, it could be on acceleration)
  * TODO check if this issue https://github.com/introlab/rtabmap/issues/1281#issuecomment-2131514123 also happens here
- [x] TODO : add LAS export option. https://github.com/introlab/rtabmap/issues/1298
  * Currently blocked by GDAL library `libgdal.a` missing after building it. Maybe try https://github.com/BeinerChes/gdalios and see what are the differences.

~~Biggest issue right now~~: not sure if it was because of new iOS updates, but ArKit does now a lot of relocalizations. In 2020, ARkit felt more like a VIO than VSLAM, so the integration with RTAB-Map was pretty smooth and worked well. The new option to filter localization jumps is inspired from ARCore workaround we had to do in the past for similar issue. In both ARKit and ARCore, there is no option to disable re-localization...

Update: The PR is ready for QA:
- [x] Google Tango
- [x] ARCore NDK
- [x] ARCore Java
- [x] AREngine?
- [x] ARKit

New issues:
 - [x] The map is rotating in yaw on loop closures (saw on iOS and android, doublecheck with tango). It looks like the [prior with minimum yaw variance](https://github.com/introlab/rtabmap/blob/538bc681d4a382ad06333ee1e17c6ac120e82c80/corelib/src/optimizer/OptimizerGTSAM.cpp#L220) on the rootid is not working.
     * Update: It seems related to gravity links, if they are not aligned with Z. When holding the phone in landscape so that pose up axis is z+ instead of y+, the yaw jumps don't happen anymore. Only affecting GTSAM optimizer, g2o looks fine. That seems also the cause of many rejected loop closures because graph optimization error is too high, then if we rotate the phone to have z+ axis up, loop closures are accepted.
     image. 
    * Decision: we will keep as "Won't fix". I looked if setting g2o by default would be better, but in multi-session mapping gtsam is better (on maps I tested). Also, the yaw jumps would appear only on loop closures, and it is "annoying" only when getting a lot of consecutive loop closures, which doesn't happen often in real mapping session (we loop once, then move exploring more). The workaround is to turn the phone so that z+ axis (blue one in third person view) is up. I also updated the noise model for gtsam gravity constraints, which seems to help a little.
- [x] On iOS, move "New Data Recording" to an Advanced menu (to avoid misclick and don't confuse people which one to use).


matlabbe:
About ARKit jumps, here is a plot of the recorded poses (without any correction, using new Data recording mode at 30 Hz):
Screenshot 2024-09-14 at 3 33 37 PM

If we zoom in on the highest velocity peak:
Screenshot 2024-09-14 at 3 35 37 PM

Here are the two corresponding point clouds on that jump:
Screen Recording 2024-09-14 at 3 57 34 PM

What it should be:
Screen Recording 2024-09-14 at 3 59 30 PM

When applying a filtering at maximum velocity of 1 m/s for minimum jump > 2 cm, we can recover a smoother odometry trajectory:
 
Screenshot 2024-09-14 at 3 39 48 PM
Screenshot 2024-09-14 at 3 41 24 PM

Note that there are still some acceleration peaks, but they are more related to noise than re-localization jumps. Here is a zoom around a noisy area:
Screenshot 2024-09-14 at 3 37 30 PM

APOS80:
This sounds realy good!

PR #1344, opened 2024-09-17
Clicking Nodes\Links brings them up in viewer

Nosille:
I added the ability to mouse click on nodes and links in the graph view and have them populate slider_A and slider_B in the main view window.

### DIFF
--- a/guilib/include/rtabmap/gui/DatabaseViewer.h
+++ b/guilib/include/rtabmap/gui/DatabaseViewer.h
@@ -52,6 +52,7 @@ class QGraphicsView;
 class QLabel;
 class QToolButton;
 class QDialog;
+class QSpinBox;
 
 namespace rtabmap
 {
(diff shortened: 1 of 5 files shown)


matlabbe:
That feature already existed here:
https://github.com/introlab/rtabmap/blob/ddb9b546ace55ca0bdf53fab3cfc9f7bcc579262/guilib/src/GraphViewer.cpp#L1915-L1939

but it was on double-click. I tried your PR and it feels better with single-click. I modified the PR to use single-click now. I also kept your spin boxes next to the main slider bars, which are indeed useful. I merged the spin box with the index label, as they represented the same function:
 
Screenshot from 2024-09-21 17-39-48

I removed the code about "highlighting" option, which seemed not used anywhere. The border context menu option also crashed when clicking on it, so I removed it. I suggest to open another MR if there was a new feature there that could be useful to keep.

I'll merge now with those changes.

PR #1351, opened 2024-10-06
Update iOS app's install_deps.sh to remove PCL's SSE dependency

tomaja:
Hi, I tried building the ios demo application and install_deps.sh was failing while trying to build PCL.
CMake was failing to produce XCode project because the check for SSE support was failing. I removed it, since as far as I know, it's not required for aarch64 builds.
Project builds nicely after updating the script.


matlabbe:
That solution is better than my workaround in my other PR (https://github.com/introlab/rtabmap/pull/1340, calling a second time cmake), I'll merge to use your approach. Thanks!

PR #1364, opened 2024-10-23
Sort keypoints before SSC

borongyuan:
I missed an important step. In ANMS, keypoints need to be sorted by strength first. When calling util2d::SSC(), keypoints can now be sorted, or an additional sorted indx can be used.


matlabbe:
LGTM!

PR #1373, opened 2024-11-02
Draw "1 - Loop Thr" along with posterior

borongyuan:
Previously, the loop threshold was drawn in the posterior plot. But I don't understand how to compare it with the posterior curve. What we can compare directly is the virtual place hypothesis and "1 - loop threshold". These changes only affect the plot.


matlabbe:
Agreed, that is easier to interpret. If first hypothesis (new place) is below the threshold, then a loop closure is triggered. I think I just never re-adjusted that line from the original code (back in 2010-2011 when the code was hosted on google code there was no virtual place hypothesis). 
image

Thanks for the update!

PR #1381, opened 2024-11-13
Concise layout

Nosille:
I run multiple cameras and as a result frequently want the images and graph in database viewer to be bigger.  These changes enable a mode that hides parts of the interface to make that true.


Nosille:
I used VerticalLayout as a template, and it included definitions in the PreferencesDialog.  I assume that without the entries in PreferencesDialog it will not save the configuration between sessions.  I will see what happens when I remove it and get back to you.

matlabbe:
The config in Preferences affects the layout under MainWindow. The config in DatabaseViewer is saved on its own.

matlabbe:
I tested it and we don't need the changes from Preferences to save/reload the config. I removed those changes. I updated the PR to not draw matching lines and to be able to click on graph view's nodes to always change graphics view A.

Thanks for the PR, I like this view mode.

PR #1405, opened 2024-12-10
Support Ceres 2.1+ Manifolds in backwards-compatible manner

ahojnnes:
Ceres 2.1+ introduces "manifolds" in replacement of "local parameterizations". This PR supports manifolds in a backwards-compatible way, such that rtabmap can be compiled with both Ceres 2.1+ or older.


matlabbe:
Great , thank you!

Issue #1362, opened 2024-10-21
[Feature-Request] RVL Depth Image Compression


borongyuan:
[RVL](https://www.microsoft.com/en-us/research/publication/fast-lossless-depth-image-compression/) is a lossless depth image compression algorithm developed by Microsoft. RVL frame is supported by some Orbbec cameras. [RVL codec](https://github.com/ros-perception/image_transport_plugins/blob/jazzy/compressed_depth_image_transport/src/rvl_codec.cpp) has been added to ROS since Noetic/Jazzy. I think it will help reduce the size of the database and save bandwidth (for camera data transfer and ROS communication).

matlabbe:
Thanks for the link, that could be indeed useful! As a comparison, we are currently using PNG:
![image](https://github.com/user-attachments/assets/6c150b6b-44ee-485e-8899-673fe4ae66d2)

Implementation details:
* Example RVL encoding: https://github.com/ros-perception/image_transport_plugins/blob/ba6a7632613ba19a6d9f3cb33483af053e6ae7a9/compressed_depth_image_transport/src/codec.cpp#L307-L319
* Example RVL decoding: https://github.com/ros-perception/image_transport_plugins/blob/ba6a7632613ba19a6d9f3cb33483af053e6ae7a9/compressed_depth_image_transport/src/codec.cpp#L113-L142
* With current rtabmap compression code:
   * on [encoding phase](https://github.com/introlab/rtabmap/blob/e3fde4ecdd8e40c6634903b6aee2cdf4b95cd24a/corelib/src/Compression.cpp#L100-L117), we could provide "rvl" format option
   * on [decoding phase](https://github.com/introlab/rtabmap/blob/e3fde4ecdd8e40c6634903b6aee2cdf4b95cd24a/corelib/src/Compression.cpp#L130-L150), we current rely on opencv decode function to know image size and type of the compressed data (which can be RGB, depth 16bits or depth 32bits). We may check if opencv is adding some header in compressed data when encoding, so we could create a new encoding header to know the compressed data is "rvl", and not "png" or "jpg". That should be backward compatible to still load old databases.

borongyuan:
Hi, RVL only supports the depth image in 16 bits format. So I went to see how the depth image in 32 bits format was processed previously. According to
https://github.com/introlab/rtabmap/blob/b2e26e81c14bd7a1394da10998d0264968089cbe/corelib/src/Compression.cpp#L105C1-L110C4
[...]

PR #1409, opened 2024-12-11
add support for RVL depth image compression

borongyuan:
addressing #1362

### DIFF
--- a/corelib/include/rtabmap/core/Compression.h
+++ b/corelib/include/rtabmap/core/Compression.h
@@ -87,5 +87,9 @@ cv::Mat RTABMAP_CORE_EXPORT uncompressData(const unsigned char * bytes, unsigned
 cv::Mat RTABMAP_CORE_EXPORT compressString(const std::string & str);
 std::string RTABMAP_CORE_EXPORT uncompressString(const cv::Mat & bytes);
 
+std::string compressedDepthFormat(const cv::Mat & bytes);
+std::string compressedDepthFormat(const std::vector<unsigned char> & bytes);
+std::string compressedDepthFormat(const unsigned char * bytes, unsigned long size);
+
 } /* namespace rtabmap */
 #endif /* COMPRESSION_H_ */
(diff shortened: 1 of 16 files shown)


borongyuan:
Compared with ROS's compressed_depth_image_transport, we add an additional signature, but there is no compressionConfig part. Therefore, additional processing is required to convert from compressedDepth message to RTAB-Map.

https://github.com/ros-perception/image_transport_plugins/blob/ccb2e9bf447c16c1857afc0d935413dcc2969cea/compressed_depth_image_transport/src/codec.cpp#L383-L392

If someone uses rtabmap_util/rgbd_relay for data compression and then rtabmap_util/rgbd_split for splitting, they will not get a valid compressedDepth message.

matlabbe:
That PR looks great! I'll check tomorrow how it could work with ROS. As you said, we may have to strip the header "DEPTHRVL" from the byte array when converting back to `sensor_msgs/ImageCompressed`. I guess for the input compressed depth images, image_transport would already convert it.

matlabbe:
On `rtabmap_ros`, we don't actually convert back from rtabmap's internal compressed depth format to the outside. For `rgbd_split`, `rgbd_relay` and `rgbd_sync` nodes, they already explicitly use PNG by default when `compress` option is enabled. So we should not break anything merging this PR.

I did a small comparison of the compression time using ZED 2K depth images:
 * RVL ~12 ms 
 * PNG 16bits: ~24 ms 
 * PNG 32 bits: ~50 ms

My main changes are:
 * In DBViewer, there is an action to edit the depth image. I added a function to read the compression format so we can re-save the edited depth to database with same format.
 * Because default of `Mem/SaveDepth16Format=false` and `Mem/DepthCompressionFormat=".rvl"`, I removed the warning about setting `Mem/SaveDepth16Format` to true if `Mem/DepthCompressionFormat=".rvl"` (as it would appear every time). If the input depth images are already 16bits, no warning will be shown and RVL is used. If the input images are 32bits, there will be a warning if RVL is used that depth images are incompatible and they will be saved in PNG instead. This will make sure that stereo depth images, that have often values over 65 meters, are not clipped by default.

PR #1412, opened 2024-12-12
Nosille small bug fix

Nosille:
I stumbled across a small indexing bug in adjustNormalsToViewPoint.  I also noted a piece of unreachable code in LocalGridMaker.

### DIFF
--- a/corelib/src/util3d_surface.cpp
+++ b/corelib/src/util3d_surface.cpp
@@ -3530,7 +3530,7 @@ LaserScan adjustNormalsToViewPoint(
 
 					float result = v.dot(n);
 					if(result < 0
-					 || (groundNormalsUp>0.0f && ptr[nz] < -groundNormalsUp && ptr[2] < viewpoint[3])) // some far velodyne rays on road can have normals toward ground
+					 || (groundNormalsUp>0.0f && ptr[nz] < -groundNormalsUp && ptr[2] < viewpoint[2])) // some far velodyne rays on road can have normals toward ground
 					{
 						//reverse normal
 						ptr[nx] *= -1.0f;


corelib/src/LocalGridMaker.cpp
```diff
@@ -258,26 +258,7 @@ void LocalGridMaker::createLocalMap(
...
-
 				createLocalMap(scan, node.getPose(), groundCells, obstacleCells, emptyCells, viewPoint);
-
-				if(scan.is2d())
-				{
-					// restore
-					normalsSegmentation_ = normalSegmentationTmp;
-					minGroundHeight_ = minGroundHeightTmp;
-					maxGroundHeight_ = maxGroundHeightTmp;
-				}
```
matlabbe: This is done in a special case when we combine 2D lidar data and 3D depth data into the local grid. Is it really never called even in that case?

corelib/src/util3d_surface.cpp
```diff
@@ -3530,7 +3530,7 @@ LaserScan adjustNormalsToViewPoint(
 
 					float result = v.dot(n);
 					if(result < 0
-					 || (groundNormalsUp>0.0f && ptr[nz] < -groundNormalsUp && ptr[2] < viewpoint[3])) // some far velodyne rays on road can have normals toward ground
```
matlabbe, line 3533: viewpoint is an Eigen::Vector3f object, with indexes 0=x, 1=y and 2=z. 3 would be out of bounds.

corelib/src/LocalGridMaker.cpp
Nosille: It grabbed my attention because of multiple checks for is2d() and occupancy == 0.  However, after a second look, you are correct that occupancy == 2 could still get there.    

My bad.

corelib/src/util3d_surface.cpp
Nosille: After, sending this I dug a little deeper and noted that there are several other places in the code where viewpoint[3] is used.  The difference is that in those cases the vector is an Eigen::Vector4f, so it doesn't crash like this did. However, it seems to me that viewpoint[2] is likely the intended array value in those cases as well.  Do you agree?  Or am a missing something?  It seems like a simple comparison between point.z and viewpoint.z, but I guess you could be doing something I am not following with the 4th value in the homogeneous coordinates.

matlabbe: Thanks for the comment, the other places using viewpoint[3] are wrong. Here the ones I've found:
https://github.com/search?q=repo%3Aintrolab%2Frtabmap%20viewpoint%5B3%5D&type=code

I created an issue about it https://github.com/introlab/rtabmap/issues/1413

PR #1437, opened 2025-01-16
fix build with ceres<2.1

borongyuan:
Related to #1405
corelib/src/optimizer/ceres/pose_graph_3d/eigen_quaternion_parameterization.h is also not compatible. However this file is not used, so I left it as is.


matlabbe:
Thanks!

PR #1445, opened 2025-01-30
fixed ORB_SLAM preproc breaking gui conditionals

laurence-diack-pk:
setting RTABMAP_ORB_SLAM true would put the if statements all out of whack in the perferences dialog:

`/home/grover/vision_dev/rtabmap/guilib/src/PreferencesDialog.cpp: In member function ‘void rtabmap::PreferencesDialog::setParameter(const string&, const string&)’:
/home/grover/vision_dev/rtabmap/guilib/src/PreferencesDialog.cpp:4904:7: error: expected ‘}’ before ‘else’
 4904 |       else
      |       ^~~~
/home/grover/vision_dev/rtabmap/guilib/src/PreferencesDialog.cpp:4894:6: note: to match this ‘{’
 4894 |      {
      |      ^
`

There may be a cleaner way to do it, but his fixed my build

### DIFF
--- a/guilib/src/PreferencesDialog.cpp
+++ b/guilib/src/PreferencesDialog.cpp
@@ -4892,10 +4892,13 @@ void PreferencesDialog::setParameter(const std::string & key, const std::string 
 				{
 					if(valueInt==2 && combo->objectName().toStdString().compare(Parameters::kOptimizerStrategy()) == 0)
 					{
+					if(
 #ifndef RTABMAP_ORB_SLAM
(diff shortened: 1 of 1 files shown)


matlabbe:
Looks like that commit https://github.com/introlab/rtabmap/commit/97c56361f8cb5bbabb401bd5ca03f4a3f27a9d2f#diff-7282862b01571b1485f94b60c0e593f502fff1cd2d4fe04c14ed633225242cdb broke it. Thanks for the fix!

PR #1446, opened 2025-02-05
RAM: add entilement com.apple.developer.kernel.increased-memory-limit…

zeitlerv:
… to increase the available ram for the app. ( Effect on iphone 12 = 3GB ->4GB)

available Memory can be checked with:

let availableMem = os_proc_available_memory()
print("Available memory: (availableMem) bytes")

and #include <os/proc.h> in Bridging-Header

### DIFF
--- a/app/ios/RTABMapApp.xcodeproj/project.pbxproj
+++ b/app/ios/RTABMapApp.xcodeproj/project.pbxproj
@@ -155,6 +155,7 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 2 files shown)


zeitlerv:
https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.kernel.increased-memory-limit

matlabbe:
That sounds great. I'll try to test it on my ipad soon.

zeitlerv:
Effect on iphone 15 Pro = 6 GB available RAM for the App instead of 3GB

Issue #1407, opened 2024-12-10
IOS: Request: More optional values in “Max features extracted”


APOS80:
Hello again.

I think 1000 is to small and infinite is to high, I would like steps up to 3000 if possible.

borongyuan:
Using too many keypoints clustered together does not actually provide more information. So I usually only use hundreds of high-quality keypoints.

matlabbe:
Re-opening, as it is a small change to do.

matlabbe:
The new version with this change has been released to App Store.

Issue #1401, opened 2024-12-07
iOS/android with TOF/LiDAR should use Vis/PnPVarianceMedianRatio=2


matlabbe:
Because it is not stereo data, the variance computation should divide only by 2. Currently the linear variance would be underestimated, causing some upcoming good loop closures to be rejected (by loop closure links, not odometry links).

APOS80:
I hope to se a fix of the app soon.

It would really help.

matlabbe:
The new version with this change has been released to App Store.

Issue #1374, opened 2024-11-04
IOS: Possibility of Creating LAZ Files on iPhone?


zeitlerv:
Hello,

I'm working on a project that requires creating LAZ files directly on an iPhone, and I was wondering if you could provide any guidance on achieving this. I’ve encountered some challenges with existing libraries and would appreciate any suggestions or pointers you might have.

Here’s what I’ve tried so far:

LibLas: Unfortunately, LibLas only supports LAS up to version 1.3, which doesn’t meet my needs for LAZ file creation.

LasZip: I've managed to build it, but it only generates .dylib files, which aren’t easily compatible with iOS. Ideally, I would need a static library that could integrate smoothly with an iOS project.

PDAL: I also attempted to build a static library with PDAL, but I was unable to produce a version that works on iOS.

If you know of any solutions or could point me in the right direction for enabling LAZ file creation on iOS, I would be extremely grateful!

Thank you for your time and any help you can provide.

Best regards,
Viktor

matlabbe:
I tried building PDAL for iOS and got many issues with the dependencies. On main dependency is gdal, you may try this repo: https://github.com/BeinerChes/gdalios, though I could not make it work on my system.

LibLAS worked, though I didn't try building it with laszip dependency.
https://github.com/introlab/rtabmap/blob/75ac722996ca13e3c2443a50ef2b00e551d9e22c/app/ios/RTABMapApp/install_deps.sh#L225-L240

It [seems 2.0.1 is required](https://github.com/libLAS/libLAS/blob/62f3d57215ba1b074a84ff08a1539b69e69e38fb/CMakeLists.txt#L201-L211) with liblas: https://github.com/LASzip/LASzip/tree/2.0.1 . Maybe it can be compiled in a similar way than liblas above, then enable it as dependency to liblas to use it.

matlabbe:
The new version with this new feature has been released to App Store.

Go under Settings->Assembling..., scroll all way down and then you can change the format of point cloud exported:

![Image](https://github.com/user-attachments/assets/0cf13315-41eb-47c1-8277-2dc91c0267bd)

zeitlerv:
@matlabbe wow, well done!

Issue #1358, opened 2024-10-19
Texturing untextured surfaces with nearby Colors within the iOS App


electrum-bowie:
![image](https://github.com/user-attachments/assets/46a26533-c96f-4e83-9fe0-93cbba64fad1)
![image](https://github.com/user-attachments/assets/ca356430-4b93-4bb7-8cbc-67cfe8ef3dca)

The vertex mesh assembling mode does this but unfortunately the textured parts are extremely low quality, and as you may know, not having textures on the model itself makes it incompatible with basic software.

electrum-bowie:
Also, if you add something like this, please keep the mesh enclosing feature ( hole filling functionality ) available, RTABMap is the only iOS software that provides actual complete 360 models, which even closes bigger voids in the capture that the LiDAR Scanner didn't reach !

It would be amazing to have a setting that can maximize the hole filling algorithm strength in the future as well, knowing that it can still leave some big gaps open. 

That combined with the requested feature in this post, additionally texturing meshes based on nearby colors, would finally enable full environmental scanning directly from the iPhone !

It would finally allow revisiting complete captures in VR, without having colorless or big voids in the models !

matlabbe:
Thanks for the suggestion. I'll check we can do that for next ios version release.

electrum-bowie:
Yup, using the PC to export a scan from a standalone device ( iOS ) to another standalone device ( Meta Quest VR ) isn't feasible.

electrum-bowie:
Does the new version include this feature ?

matlabbe:
Not yet.

DenisDaniel707:
This feature would be a game changer for sure! The app is already a powerhouse. Hats off to @matlabbe 🙏

matlabbe:
To add to the discussion, it seems that we can add vertex color in OBJ format, though it is not standard, so not sure about the support in software you may use. Ref: https://paulbourke.net/dataformats/obj/colour.html

If it works, that would be the easiest/fastest way to do it, as we should already know the closest color of each vertex in the mesh before texturing it.

matlabbe:
The new version with this new feature has been released to App Store.

You can enable this export mode by changing the following settings under Settings->Assembling:

![Image](https://github.com/user-attachments/assets/4be625ff-1ddc-4cbc-b03b-602162ad519b)

This makes the mesh fully closed:
![Image](https://github.com/user-attachments/assets/9efd5071-5eb1-411e-aa74-9436fd665911)

[...]

PR #1451, opened 2025-02-16
iOS feb 2025 various updates

matlabbe:
Fixes:
- [x] https://github.com/introlab/rtabmap/issues/1401
- [x] https://github.com/introlab/rtabmap/issues/1407
- [x] https://github.com/introlab/rtabmap/issues/1455
- [x] https://github.com/introlab/rtabmap/issues/1374

New features:
- [x] Measuring tool
- [x] https://github.com/introlab/rtabmap/issues/1358
    * Currently implemented like Meshlab is rendering, so we see seams between texture and coloured vertex. That may be useful or not, maybe add a rendering option to not show them. This can be done in the fragment shader, by only applying color when texcoord >= 0.99

Review: 
- [x] The measuring button shows up when visualizing optimized point cloud, but cannot be used. Hide button if we are visualizing a point cloud.
- [x] When export in LAZ format, we shouldn't zip the file afterwards, it won't be smaller anyway. Just export directly as laz instead of zip.


matlabbe:
The app has been submitted to App Store.

hkortier:
Building for iOS yields two issues:
- [No matching function for call to 'mergeTextures'](https://github.com/introlab/rtabmap/blob/26b28e1360eee0332df27d33f6de06d72962bd49/app/android/jni/RTABMapApp.cpp#L3813) 
- [No member named 'saveOBJFile' in namespace 'rtabmap::util3d'; did you mean 'pcl::io::saveOBJFile'?](https://github.com/introlab/rtabmap/blob/26b28e1360eee0332df27d33f6de06d72962bd49/app/android/jni/RTABMapApp.cpp#L4290)

matlabbe:
@hkortier Make sure you rebuild the rtabmap library:
https://github.com/introlab/rtabmap/blob/26b28e1360eee0332df27d33f6de06d72962bd49/app/ios/RTABMapApp/install_deps.sh#L269-L271

PR #1460, opened 2025-03-04
Improved depth estimation of mono features

matlabbe:
Added parameters:
* `OdomF2M/BundleUpdateFeatureMapOnAllFrames`: Update 3D local feature map on every frame with bundle adjustment. Recommended if `Vis/DepthAsMask=false` and `Mem/UseOdomFeatures=true` so that features without depth are better triangulated on every frame (not only on keyframes). If disabled, the feature map is updated only when a new keyframe is added (legacy approach).
* `OdomF2M/FloorThreshold`: Only track features in 3D feature map that are over this threshold (height in base frame). Can be useful to ignore reflections on the floor. 0 means disabled.
* `OdomF2M/InitDepthFactor`: Depth factor used to initialize depth of features without depth. Depth = Factor * fx.

TODO: Re-run vslam benchmark datasets with and without `OdomF2M/BundleUpdateFeatureMapOnAllFrames` to see if it doesn't affect too negatively the results. Ideally, we would want `OdomF2M/BundleUpdateFeatureMapOnAllFrames` to be true by default so that depth of mono features can be correctly estimated all the time.

### DIFF
--- a/corelib/src/odometry/OdometryF2M.cpp
+++ b/corelib/src/odometry/OdometryF2M.cpp
@@ -60,6 +60,7 @@ OdometryF2M::OdometryF2M(const ParametersMap & parameters) :
 	keyFrameThr_(Parameters::defaultOdomKeyFrameThr()),
 	visKeyFrameThr_(Parameters::defaultOdomVisKeyFrameThr()),
 	maxNewFeatures_(Parameters::defaultOdomF2MMaxNewFeatures()),
+	floorThreshold_(Parameters::defaultOdomF2MFloorThreshold()),
 	scanKeyFrameThr_(Parameters::defaultOdomScanKeyFrameThr()),
 	scanMaximumMapSize_(Parameters::defaultOdomF2MScanMaxSize()),
 	scanSubtractRadius_(Parameters::defaultOdomF2MScanSubtractRadius()),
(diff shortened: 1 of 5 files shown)


matlabbe:
Euroc datasets are the more negatively impacted by `OdomF2M/BundleUpdateFeatureMapOnAllFrames=true`, we should keep it to false by default.

corelib/src/odometry/OdometryF2M.cpp
```diff
@@ -818,6 +870,28 @@ Transform OdometryF2M::computeTransform(
...
+								float y = iter->second.second.first.pt.y;
+								Eigen::Vector3f ray = util3d::projectDepthTo3DRay(
+										lastFrameModels[cameraIndex].imageSize(),
+										x,
+										y,
+										lastFrameModels[cameraIndex].cx(),
+										lastFrameModels[cameraIndex].cy(),
+										lastFrameModels[cameraIndex].fx(),
+										lastFrameModels[cameraIndex].fy());
+								float scaleInf = 0.05 * lastFrameModels[cameraIndex].fx();
```
matlabbe: Doublecheck how much this affects the results. Maybe Euroc dataset didn't work well because of this change.

matlabbe: The dataset evaluation is using `Vis/DepthAsMask=true` so this is ignored anyway. I still added a parameter called `OdomF2M/InitDepthFactor` to tune 0.05 if needed.

matlabbe:
I tested again Euroc with live visualization and the results at 1 Hz are pretty close to original results. It seems that a not so accurate loop closure "sometimes" happens, which add significant error, though if I remove that link, the results are close to paper.

I'll still keep `OdomF2M/BundleUpdateFeatureMapOnAllFrames=false` by default till I have more real experiments with `OdomF2M/BundleUpdateFeatureMapOnAllFrames=true`.

PR #1470, opened 2025-03-17
Enable on-device feature detectors for color camera

borongyuan:
On-device feature detectors can use color camera input. I forget why I thought this wasn't available before. It seemed to be due to some performance issues. But my current test only found that using GFTT in high-resolution configuration may not achieve 30 FPS.


matlabbe:
Understood, thanks for the PR!

PR #1504, opened 2025-05-13
Add OpenCL GPU implementation

sssarana:
- add dependency on `grid_map_core`;
- fix build with NONFREE algorithms for `humble-devel`;
- Replace Mat objects with UMat to use with OpenCL;

This will automatically use OpenCL GPU if it is available.

### DIFF
--- a/corelib/src/Features2d.cpp
+++ b/corelib/src/Features2d.cpp
@@ -1140,6 +1140,7 @@ void SURF::parseParameters(const ParametersMap & parameters)
 	{
 #if CV_MAJOR_VERSION < 3
 		_surf = cv::Ptr<CV_SURF>(new CV_SURF(hessianThreshold_, nOctaves_, nOctaveLayers_, extended_, upright_));
+		std::cout << "using cpu" << std::endl;
 #else
 		_surf = CV_SURF::create(hessianThreshold_, nOctaves_, nOctaveLayers_, extended_, upright_);
 #endif
--- a/corelib/src/Features2d.cpp
+++ b/corelib/src/Features2d.cpp
@@ -39,6 +39,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <opencv2/imgproc/imgproc_c.h>
 #include <opencv2/core/version.hpp>
 #include <opencv2/opencv_modules.hpp>
+#include <opencv2/core/ocl.hpp>
 
 #ifdef RTABMAP_ORB_OCTREE
 #include "opencv/ORBextractor.h"
@@ -769,13 +770,16 @@ std::vector<cv::KeyPoint> Feature2D::generateKeypoints(const cv::Mat & image, co
 	UASSERT(!image.empty());
 	UASSERT(image.type() == CV_8UC1);
 
-	cv::Mat mask;
+	cv::UMat imageU;
+    image.copyTo(imageU);
+	cv::UMat maskU;
+	cv::Mat maskMat;
 	if(!maskIn.empty())
 	{
 		if(maskIn.type()==CV_16UC1 || maskIn.type() == CV_32FC1)
 		{
 			mask = cv::Mat::zeros(maskIn.rows, maskIn.cols, CV_8UC1);
-			for(int i=0; i<(int)mask.total(); ++i)
+			for(int i=0; i<(int)maskMat.total(); ++i)
 			{
 				float value = 0.0f;
 				if(maskIn.type()==CV_16UC1)
@@ -795,14 +799,15 @@ std::vector<cv::KeyPoint> Feature2D::generateKeypoints(const cv::Mat & image, co
 				   (_maxDepth == 0.0f || value <= _maxDepth) &&
 				   uIsFinite(value))
 				{
-					((unsigned char*)mask.data)[i] = 255; // ORB uses 255 to handle pyramids
+					maskMat.data[i] = 255;
 				}
 			}
+			maskMat.copyTo(maskU);
 		}
 		else if(maskIn.type()==CV_8UC1)
 		{
 			// assume a standard mask
-			mask = maskIn;
+			maskIn.copyTo(maskU);
 		}
 		else
 		{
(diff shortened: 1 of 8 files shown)


corelib/src/Features2d.cpp
```diff
@@ -1140,6 +1145,7 @@ void SURF::parseParameters(const ParametersMap & parameters)
 	{
 #if CV_MAJOR_VERSION < 3
 		_surf = cv::Ptr<CV_SURF>(new CV_SURF(hessianThreshold_, nOctaves_, nOctaveLayers_, extended_, upright_));
+		std::cout << "using cpu" << std::endl;
```
matlabbe: That log could be removed.

corelib/src/Features2d.cpp
```diff
@@ -1172,11 +1178,20 @@ std::vector<cv::KeyPoint> SURF::generateKeypointsImpl(const cv::Mat & image, con
...
+        imgRoi.copyTo(imgRoiU);
+
+        cv::UMat maskRoiU;
+        if (!maskRoi.empty())
+        {
+            maskRoi.copyTo(maskRoiU);
+        }
+
+        _surf->detect(imgRoiU, keypoints, maskRoiU);
+    }
```
matlabbe, line 1193: Would we need to do something similar for all detectors? I may create an issue to do it later for the others.

corelib/src/MarkerDetector.cpp
```diff
@@ -246,11 +247,13 @@ std::map<int, MarkerInfo> MarkerDetector::detect(const cv::Mat & image,
             std::map<int, float>::const_iterator findIter = markerLengths.find(ids[i]);
             if(!depth.empty() && (markerLength_ == 0 || (markerLength_<0 && findIter==markerLengths.end())))
 			{
-                float d = util2d::getDepth(depth, (corners[i][0].x + (corners[i][2].x-corners[i][0].x)/2.0f)*rgbToDepthFactorX, (corners[i][0].y + (corners[i][2].y-corners[i][0].y)/2.0f)*rgbToDepthFactorY, true, 0.02f, true);
-				float d1 = util2d::getDepth(depth, corners[i][0].x*rgbToDepthFactorX, corners[i][0].y*rgbToDepthFactorY, true, 0.02f, true);
-				float d2 = util2d::getDepth(depth, corners[i][1].x*rgbToDepthFactorX, corners[i][1].y*rgbToDepthFactorY, true, 0.02f, true);
-				float d3 = util2d::getDepth(depth, corners[i][2].x*rgbToDepthFactorX, corners[i][2].y*rgbToDepthFactorY, true, 0.02f, true);
-				float d4 = util2d::getDepth(depth, corners[i][3].x*rgbToDepthFactorX, corners[i][3].y*rgbToDepthFactorY, true, 0.02f, true);
+				cv::Mat depthMat;
+				depth.copyTo(depthMat);
```
matlabbe:
```suggestion
				depthMat = depth.getMat(cv::ACCESS_READ);

```
copy only if needed

corelib/src/MarkerDetector.cpp
```diff
@@ -363,9 +366,12 @@ std::map<int, MarkerInfo> MarkerDetector::detect(const cv::Mat & image,
 
 	if(imageWithDetections)
 	{
+		cv::Mat detections_img;
+		imageWithDetections->copyTo(detections_img);
+		UWARN("entering this part");
```
matlabbe: Log to remove.

corelib/src/odometry/OdometryF2M.cpp
```diff
@@ -1263,6 +1266,7 @@ Transform OdometryF2M::computeTransform(
 
 					map_->setWords(words, wordsKpts, transformedPoints, descriptors);
 					addKeyFrame = true;
+					descriptors.release();
```
matlabbe: Is it necessary?

package.xml
```diff
@@ -22,7 +21,7 @@
   <!-- <depend>libproj-dev</depend> needed due to error in vtk6 (kinetic)-->
   <depend>libsqlite3-dev</depend>
   <depend>liboctomap-dev</depend>
-  <!-- <depend>grid_map_core</depend> --> <!-- till this PR is released https://github.com/ANYbotics/grid_map/pull/499 -->
+  <depend>grid_map_core</depend> <!-- till this PR is released https://github.com/ANYbotics/grid_map/pull/499 -->
```
matlabbe: I think we still need to wait for https://github.com/ANYbotics/grid_map/pull/519 to be released, we may change the comment to 
```suggestion
  <!-- <depend>grid_map_core</depend> --> <!-- till this PR is released https://github.com/ANYbotics/grid_map/pull/519 -->
```
Note that dependency in `package.xml` is useful only if you use rosdep to install dependencies, the workaround if you want `grid_map_core` support locally is to install explicitly `grid_map_core` and cmake should be able to find it.

matlabbe:
I kinda forgot about `UMat` when OpenCV3 came out, thinking it was already integrated inside `cv::Mat`, though it should be explicitly used. I put some comments above. It looks like for most functions, we would need to convert to `cv::InputArray` / `cv::OutputArray` to avoid unnecessary copies between GPU and CPU. That would be a bigger PR though. I'll try to take time to test at least your changes to see what is the impact on those functions (using OpenCL or not) compared to old code, then merge it for now (and do UMat support for all other detectors later).

corelib/src/Features2d.cpp
sssarana: Yeah, all detectors should have similar implementation if we want to use them on GPU. It would be great to have it.

corelib/src/MarkerDetector.cpp
sssarana: The `depthMat = depth.getMat(cv::ACCESS_READ);` caused some lifecycle issues for me earlier, so I wouldn't trust it. The reason for that is the original object that we create a shallow copy from gets deleted, so the `depthMat` gets deleted also.

package.xml
sssarana: I created a PR to fix `grid_map_core` here: ANYbotics/grid_map#521, but it didn't get reviewed yet. Though I think you can apply these fixes yourself if you want to test everything properly.

sssarana:
Thanks for your feedback. I fixed the errors and removed unnecessary logs now, let me know if anything else is needed.
Using the `cv::InputArray`/`cv::OutputArray` is a good idea, it will also make implementation for `UMat` support on other detectors a bit easier. 
Thanks.

matlabbe:
I didn't have time to test the impact of these changes. In the mean time, if you can rebase this MR on `master` branch isntead of `humble-devel` it would be ideal.

PR #1535, opened 2025-06-13
Fixed a typo with SuperPoint

pzvelev:
Fixed a typo in the SuperPoint error message in Features2d.cpp where "SupertPoint" was incorrectly spelled instead of "SuperPoint".


matlabbe:
You're right, thanks for the PR!

PR #1583, opened 2025-09-17
CuVSLAM VO Strategy

Dekempsy4:
Implementing Isaac ROS' CuVSLAM visual odometry strategy as a 3rd party strategy in RTAB-Map.

cuvslam headerfile: 
https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_nitros/blob/main/isaac_ros_nitros/lib/cuvslam/include/cuvslam.h

### To run with CuVSLAM
Set the following environment variable so the system versions of `libpng16`, `libjpeg`, and `libjsoncpp` are loaded **before** cuVSLAM’s bundled copies. This avoids runtime crashes caused by conflicting dynamic libraries:

```bash
VAL="$(ldconfig -p | awk '/lib(png16|jpeg|jsoncpp)\.so/{print $NF}' | sort -u | tr '\n' ' ' | sed 's/[[:space:]]*$//')"
export LD_PRELOAD="$VAL"
```

### cuVSLAM prerequisites (Ubuntu 22.04, x86_64)

**1) NVIDIA Driver + CUDA 12.6 (toolkit)**
https://developer.nvidia.com/cuda-12-6-0-download-archive?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=deb_network

**2) NVIDIA VPI**
https://docs.nvidia.com/vpi/3.0/installation.html

**3) Isaac ROS Visual SLAM (NITROS/cuVSLAM)**
```bash
sudo apt-get update
sudo apt-get install -y ros-humble-isaac-ros-visual-slam
```


CMakeLists.txt
```diff
@@ -849,6 +850,42 @@ IF(WITH_ORB_SLAM AND NOT G2O_FOUND)
...
+            SET(cuvslam_FOUND TRUE)
+            SET(cuvslam_INCLUDE_DIRS ${CUVSLAM_INCLUDE_DIR})
+            SET(cuvslam_LIBRARIES ${CUVSLAM_LIBRARY})
+            SET(RTABMAP_CUVSLAM TRUE)
+            MESSAGE(STATUS "Found cuVSLAM: ${cuvslam_INCLUDE_DIRS}")
+        ENDIF()
+    ENDIF()
+ENDIF(WITH_CUVSLAM)
+))))))))))))))))))))))))))))
+
```
matlabbe: Could be moved into a `FindCuVSLAM.cmake` file under `cmake_modules` folder, if `cuvslam` library doesn't provide already a cmake config to find it.

corelib/include/rtabmap/core/Parameters.h
```diff
@@ -456,7 +456,7 @@ class RTABMAP_CORE_EXPORT Parameters
     RTABMAP_PARAM(GTSAM, IncRelinearizeSkip,       int, 1, "Only relinearize any variables every X calls to ISAM2::update(). See GTSAM::ISAM2 doc for more info.");
 
     // Odometry
-    RTABMAP_PARAM(Odom, Strategy,               int, 0,       "0=Frame-to-Map (F2M) 1=Frame-to-Frame (F2F) 2=Fovis 3=viso2 4=DVO-SLAM 5=ORB_SLAM2 6=OKVIS 7=LOAM 8=MSCKF_VIO 9=VINS-Fusion 10=OpenVINS 11=FLOAM 12=Open3D");
+    RTABMAP_PARAM(Odom, Strategy,               int, 0,       "0=Frame-to-Map (F2M) 1=Frame-to-Frame (F2F) 2=Fovis 3=viso2 4=DVO-SLAM 5=ORB_SLAM2 6=OKVIS 7=LOAM 8=MSCKF_VIO 9=VINS-Fusion 10=OpenVINS 11=FLOAM 12=Open3D 13=cuVSLAM");
```
matlabbe, line 459: You can use QtDesigner (or directly `$ designer` if qt is installed) and edit the odometry strategy comboBox in `preferencesDialog.ui` file.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,42 @@
+#ifndef ODOMETRYCUVSLAM_H_
+#define ODOMETRYCUVSLAM_H_
+
+#include <rtabmap/core/Odometry.h>
+#include <cuvslam/cuvslam.h>
```
matlabbe: Ideally, we would not want to have this header in the public interface of rtabmap. `#include <cuvslam/cuvslam.h>` would only be in the cpp.

If classes or functions of CuVSLAM used below can be forward declared, that would be better to do so.

See for example what we did for OdometryOpenVINS: https://github.com/introlab/rtabmap/blob/master/corelib/include/rtabmap/core/odometry/OdometryOpenVINS.h

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,42 @@
...
+        virtual void reset(const Transform & initialPose = Transform::getIdentity());
+        virtual Odometry::Type getType() {return Odometry::kTypeCuVSLAM;}
+
+    private:
+        virtual Transform computeTransform(SensorData & image, const Transform & guess = Transform(), OdometryInfo * info = 0);
+
+    private:
+    #ifdef RTABMAP_CUVSLAM
+        bool initializeCuVSLAM(const SensorData & data);
+        bool prepareImages(const SensorData & data, std::vector<CUVSLAM_Image> & cuvslam_images);
```
matlabbe: This function seems not affecting any private members, and because it is using a CUVSLAM type, declare/implement only in cpp as a free function.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,42 @@
...
+        void processIMUData(const SensorData & data);
+        Transform convertCuVSLAMPose(const CUVSLAM_Pose & cuvslam_pose);
+        CUVSLAM_Configuration CreateConfiguration(const CUVSLAM_Pose & cv_base_link_pose_cv_imu, const SensorData & data);
```
matlabbe: These functions seem not affecting any private members, and because they are using a CUVSLAM type, declare/implement only in cpp as free functions.

CMakeLists.txt
Dekempsy4: I moved these instructions to a FindCuVSLAM.cmake file.

corelib/include/rtabmap/core/Parameters.h
Dekempsy4: I've updated the ui element to include CuVSLAM

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
Dekempsy4: I removed the private memebers from the header.

cmake_modules/FindCuVSLAM.cmake
```diff
@@ -0,0 +1,71 @@
...
+#
+#  CUVSLAM_ROOT_DIR environment variable can be set to find the library.
+#
+# It sets the following variables:
+#  CUVSLAM_FOUND         - Set to false, or undefined, if cuVSLAM isn't found.
+#  CUVSLAM_INCLUDE_DIRS  - The cuVSLAM include directory.
+#  CUVSLAM_LIBRARIES     - The cuVSLAM library to link against.
+
+# Always do manual search
+if(TRUE)
```
matlabbe: You can remove this, indeed always to manual search in this file.

CMakeLists.txt
```diff
@@ -849,6 +850,20 @@ IF(WITH_ORB_SLAM AND NOT G2O_FOUND)
...
+    FIND_PACKAGE(CuVSLAM QUIET)
+    IF(CUVSLAM_FOUND)
+        MESSAGE(STATUS "Found cuVSLAM: ${CUVSLAM_INCLUDE_DIRS}")
+        # cuVSLAM requires CUDA runtime
+        find_package(CUDA REQUIRED)
+        IF(CUDA_FOUND)
+            MESSAGE(STATUS "Found CUDA for cuVSLAM: ${CUDA_INCLUDE_DIRS}")
+        ELSE()
+            MESSAGE(FATAL_ERROR "CUDA is required to build with cuVSLAM! Set -DWITH_CUVSLAM=OFF if you don't have CUDA.")
+        ENDIF()
```
matlabbe: What about making this check inside `FindCuVSLAM.cmake`?

You may either do
1. a `find_package(CUDA REQUIRED)` inside `FindCuVSLAM.cmake`, and cmake will automatically exit if CUDA is not found (because of keywork REQUIRED), or
2. a `find_package(CUDA)` (not "required" to handle error ourself) and do the `IF(CUDA_FOUND)` with a warning instead of fatal, then return CUVSLAM_FOUND=NO.

cmake_modules/FindCuVSLAM.cmake
```diff
@@ -0,0 +1,71 @@
...
+            $ENV{CUVSLAM_ROOT}/lib
+            $ENV{CUVSLAM_ROOT_DIR}/lib
+    )
+    
+    if(CUVSLAM_INCLUDE_DIRS AND CUVSLAM_LIBRARY)
+        set(CUVSLAM_FOUND TRUE)
+        set(CUVSLAM_LIBRARIES ${CUVSLAM_LIBRARY})
+    endif()
+endif()
+
```
matlabbe, line 48: Following comment above with CUDA found inside that file, you could add the `${CUDA_LIBRARIES}` and `${CUDA_INCLUDE_DIRS}` to `CUVSLAM_LIBRARIES` and `CUVSLAM_INCLUDE_DIRS` variables. So downstream, we can either do:
1. (old cmake way) 
```cmake
IF(CUVSLAM_FOUND)
    SET(INCLUDE_DIRS
		${INCLUDE_DIRS}
		${CUVSLAM_INCLUDE_DIRS}
	)
	SET(LIBRARIES
		${LIBRARIES}
		${CUVSLAM_LIBRARIES}
	)
ENDIF(CUVSLAM_FOUND)
```
2. (modern cmake way, use target) 
```cmake
IF(CUVSLAM_FOUND)
	SET(LIBRARIES
		${LIBRARIES}
		cuvslam::cuvslam
	)
ENDIF(CUVSLAM_FOUND)
```
and CUDA includes and libraries will be included indirectly.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,115 @@
+/*
+Copyright (c) 2010-2016, Mathieu Labbe - IntRoLab - Universite de Sherbrooke
```
matlabbe, line 3: You can add another line with:
```
Copyright (c) 2025 Your Name or github username
```

corelib/src/CMakeLists.txt
```diff
@@ -764,6 +765,27 @@ IF(ORB_SLAM_FOUND)
...
+		SET(INCLUDE_DIRS
+			${INCLUDE_DIRS}
+			${CUDA_INCLUDE_DIRS}
+		)
+		SET(LIBRARIES
+			${LIBRARIES}
+			${CUDA_LIBRARIES}
+		)
+	ENDIF(CUDA_FOUND)
+ENDIF(CUVSLAM_FOUND)
```
matlabbe, line 773: following the comments above, suggestion:
```cmake
IF(CUVSLAM_FOUND)
	SET(LIBRARIES
		${LIBRARIES}
		cuvslam::cuvslam
	)
ENDIF(CUVSLAM_FOUND)
```

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1070 @@
...
+#include "rtabmap/utilite/UTimer.h"
+#include "rtabmap/core/CameraModel.h"
+#include "rtabmap/core/StereoCameraModel.h"
+
+#ifdef RTABMAP_CUVSLAM
+#include <cuvslam.h>
+#include <opencv2/opencv.hpp>
+#include <tf2/LinearMath/Transform.h>
+#include <tf2/LinearMath/Matrix3x3.h>
+#include <tf2/LinearMath/Vector3.h>
```
matlabbe: You have to search for `tf2` library too inside `FindCuVSLAM.cmake` and add includes/libraries like CUDA to `cuvslam::cuvslam` cmake target

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    if(cuvslam_camera_objects_) {
+        delete cuvslam_camera_objects_;
+        cuvslam_camera_objects_ = nullptr;
+    }
+    
+    // Clean up image objects vector
+    if(cuvslam_image_objects_) {
+        delete cuvslam_image_objects_;
+        cuvslam_image_objects_ = nullptr;
+    }
```
matlabbe: Suggestion:
```
delete cuvslam_camera_objects_;
delete cuvslam_image_objects_;
```
This should work as long as `cuvslam_camera_objects_` is set to `nullptr` if already deleted.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    
+    // Clean up camera rig and configuration
+    if(camera_rig_) {
+        delete camera_rig_;
+        camera_rig_ = nullptr;
+    }
+    if(configuration_) {
+        delete configuration_;
+        configuration_ = nullptr;
+    }
```
matlabbe: Suggestion:
```
delete camera_rig_;
delete configuration_;
```
This should work as long as `camera_rig_` is set to `nullptr` if already deleted.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,115 @@
...
+	virtual Transform computeTransform(SensorData & image, const Transform & guess = Transform(), OdometryInfo * info = 0);
+
+	// Helper functions for cuVSLAM integration (internal implementation details)
+	bool initializeCuVSLAM(const SensorData & data);
+	bool prepareImages(const SensorData & data, std::vector<CUVSLAM_Image*> & cuvslam_images);
+	Transform convertCuVSLAMPose(const CUVSLAM_Pose * cuvslam_pose);
+	cv::Mat convertCuVSLAMCovariance(const float * cuvslam_covariance);
+	CUVSLAM_Configuration * CreateConfiguration(const CUVSLAM_Pose * cv_base_link_pose_cv_imu);
+	CUVSLAM_Pose * convertCameraPoseToCuVSLAM(const Transform & transform);
+	CUVSLAM_Pose * rtabmapTransformToCuVSLAMPose(const Transform & transform);
```
matlabbe: Move these forward declarations at the top of the cpp file. Remove pointers if possible.

cmake_modules/FindCuVSLAM.cmake
Dekempsy4: removed the unnecessary conditional

CMakeLists.txt
Dekempsy4: moved the find cuda logic into FindCuVSLAM.cmake

cmake_modules/FindCuVSLAM.cmake
Dekempsy4: Added new and old cmake support for FindCuda in the cuVSLAM cmake

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
Dekempsy4: Added copyright to OdometryCuVSLAM

corelib/src/CMakeLists.txt
Dekempsy4: updated this part of the cmake to reflect other changes

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: added logic for finding tf2 and eigen in the cuVSLAM cmake

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: simplified memory freeing for both gpu and cpu memory allocation.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: simplified memory freeing for both gpu and cpu memory allocation.

cmake_modules/FindCuVSLAM.cmake
```diff
@@ -0,0 +1,85 @@
...
+    HANDLE_COMPONENTS
+)
+
+if(CUVSLAM_FOUND)
+    # Create imported target for modern CMake usage
+    if(NOT TARGET cuvslam::cuvslam)
+        add_library(cuvslam::cuvslam UNKNOWN IMPORTED)
+        set_target_properties(cuvslam::cuvslam PROPERTIES
+            IMPORTED_LOCATION "${CUVSLAM_LIBRARY}"
+            INTERFACE_INCLUDE_DIRECTORIES "${CUVSLAM_INCLUDE_DIRS};${CUDA_INCLUDE_DIRS};${TF2_INCLUDE_DIRS};${EIGEN3_INCLUDE_DIR}"
```
matlabbe: Based on the code above, `CUVSLAM_INCLUDE_DIRS` may already include the other dependencies.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1085 @@
...
+        cuvslam_handle_ = nullptr;
+    }
+        
+    for(CUVSLAM_Camera * camera : cuvslam_cameras_) {
+        delete camera;
+    }
+    cuvslam_cameras_.clear();
+    
+    delete camera_rig_;
+    camera_rig_ = nullptr;
```
matlabbe: In the destructor, it doesn't really matter to reassign to null, that variable will never be used anymore... unless in the destructor logic below it is checked again -> though if so I would change the order of what is destructed to avoid this.

Check for other variables below too.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1085 @@
...
+        right_camera_params_[3] = static_cast<float>(rightModel.fy());   // fy
+        right_camera.parameters = right_camera_params_;
+        
+        // Set pinhole model with zero distortion (rectified images)
+        right_distortion_model_ = "pinhole";
+        right_camera.distortion_model = right_distortion_model_;
+        right_camera.num_parameters = 4;
+        
+        // Set right camera pose using stereo baseline (proper stereo configuration)
+        double baseline = data.stereoCameraModels()[0].baseline();
```
matlabbe:
```suggestion
        double baseline = stereoModel.baseline();
```

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,101 @@
...
+	virtual Odometry::Type getType() {return Odometry::kTypeCuVSLAM;}
+
+private:
+	virtual Transform computeTransform(SensorData & image, const Transform & guess = Transform(), OdometryInfo * info = 0);
+
+private:
+#ifdef RTABMAP_CUVSLAM
+	// cuVSLAM handles and data structures (using forward-declared types for type safety)
+	CUVSLAM_Tracker * cuvslam_handle_;
+	std::vector<CUVSLAM_Camera*> cuvslam_cameras_;
```
matlabbe: That private member `cuvslam_cameras_` could be removed. Looking at the code, we don't need to keep it after initialization. The only other place used is just to check the number of cameras, for which we can use a member just for that info if needed.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,101 @@
...
+	std::vector<CUVSLAM_Camera> * cuvslam_camera_objects_;
```
matlabbe: Same comment for this one. (not sure we need to keep this as a private member)

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,87 @@
...
+private:
+	virtual Transform computeTransform(SensorData & image, const Transform & guess = Transform(), OdometryInfo * info = 0);
+
+private:
+#ifdef RTABMAP_CUVSLAM
+	// cuVSLAM handles and data structures (using forward-declared types for type safety)
+	std::vector<CUVSLAM_Camera> * cuvslam_camera_objects_;
+	std::vector<CUVSLAM_Image> * cuvslam_image_objects_;
+	CUVSLAM_Tracker * cuvslam_handle_;
+
```
matlabbe, line 62: Note sure about this one, but the vector seems only populated when a cuvslam update is done, the vector could be removed/cleared afterwards, so maybe no need to keep it as private member either.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
Dekempsy4: haha i was actually just investigating these members. I've eliminated the unecessary private members now.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,87 @@
...
+private:
+#ifdef RTABMAP_CUVSLAM
+	// cuVSLAM handles and data structures (using forward-declared types for type safety)
+	std::vector<CUVSLAM_Camera> * cuvslam_camera_objects_;
+	std::vector<CUVSLAM_Image> * cuvslam_image_objects_;
+	CUVSLAM_Tracker * cuvslam_handle_;
+	
+	// State tracking
+	bool initialized_;
```
matlabbe, line 64: Same comment. Do we really need to keep them as private members after we initialize?

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1021 @@
...
+    // RTAB-Map order: (x, y, z, rotation about X, rotation about Y, rotation about Z)
+    Eigen::Matrix<float, 6, 6> rtabmap_covariance_mat = Eigen::Matrix<float, 6, 6>::Zero();
+    rtabmap_covariance_mat.block<3, 3>(0, 0) = covariance_mat.block<3, 3>(3, 3);  // translation-translation
+    rtabmap_covariance_mat.block<3, 3>(0, 3) = covariance_mat.block<3, 3>(3, 0);  // translation-rotation
+    rtabmap_covariance_mat.block<3, 3>(3, 0) = covariance_mat.block<3, 3>(0, 3);  // rotation-translation
+    rtabmap_covariance_mat.block<3, 3>(3, 3) = covariance_mat.block<3, 3>(0, 0);  // rotation-rotation
+    
+    // Apply coordinate system transformation
+    Eigen::Matrix<float, 6, 6> covariance_mat_change_basis = 
+        block_canonical_pose_cuvslam * rtabmap_covariance_mat * block_canonical_pose_cuvslam.transpose();
```
matlabbe, line 1002: ~Not sure why a transformation is required.~ cuvslam is using a different coordinate system that rtabmap/ros

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1021 @@
...
+tf2::Transform FromcuVSLAMPose(const CUVSLAM_Pose & cuvslam_pose)
+{
+  const auto & r = cuvslam_pose.r;
+  const auto & t = cuvslam_pose.t;
+  // tf2::Matrix3x3 is row major and cuVSLAM rotation mat is column major.
+  const tf2::Matrix3x3 rotation(r[0], r[3], r[6], r[1], r[4], r[7], r[2], r[5], r[8]);
+  const tf2::Vector3 translation(t[0], t[1], t[2]);
+
+  return tf2::Transform(rotation, translation);
+}
```
matlabbe: You could convert to `rtabmap::Transform` directly:
```
return rtabmap::Transform(r[0], r[3], r[6], t[0], r[1], r[4], r[7], t[1], r[2], r[5], r[8], t[2]);
```

EDIT: I see that is to be used with tf2 transform to change coordinate systems. One simplification could be to use `rtabmap::Transform` instead of `tf2::Transform` to remove a dependency on tf2.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
Dekempsy4: I've moved all these member functions into the cpp and made them return objects instead of pointers whenever possible.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,972 @@
...
+    if(gpu_left_image_data_) {
+        cudaFree(gpu_left_image_data_);
+    }
+    if(gpu_right_image_data_) {
+        cudaFree(gpu_right_image_data_);
+    }
+#endif
+}
+
+void OdometryCuVSLAM::reset(const Transform & initialPose)
```
matlabbe, line 248: Doublecheck that we can start/stop/start without rebooting the app if it is still an issue.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,972 @@
...
+    // Core Visual Odometry Settings 
+    configuration->multicam_mode = 0;                    // Single camera setup
+    configuration->use_motion_model = 1;                 // Enable motion model for better tracking
+    configuration->use_denoising = 0;                    // Disable denoising by default
+    configuration->use_gpu = 1;                          // Use GPU acceleration
+    configuration->horizontal_stereo_camera = 1;         // Stereo camera configuration
+    
+    // SLAM Disabled (VO-only mode) 
+    configuration->enable_localization_n_mapping = 0;    // Disable SLAM - VO only
+    configuration->enable_observations_export = 0;       // Not needed for VO
```
matlabbe: Could be interesting to enable this to see features that are used in the odometry view. You would then need to fill the `info->newCorners` member in the `OdometryInfo` object in `computeTransform()`.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: I removed the tf2 dependency, now converting directly between cuvslam (OpenGL) coords and rtabmap frames.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,940 @@
...
+    initialized_(false),
+    lost_(false),
+    previous_pose_(Transform::getIdentity()),
+    last_timestamp_(-1.0),
+    gpu_left_image_data_(nullptr),
+    gpu_right_image_data_(nullptr),
+    gpu_left_image_size_(0),
+    gpu_right_image_size_(0)
+#endif
+{
```
matlabbe, line 212: Suggestion, adding a private member `bool planar_constraints_;`
```cpp
    ...
    gpu_right_image_size_(0),
    planar_constraints_(false)
{
    Parameters::parse(parameters, Parameters::kRegForce3DoF(), planar_constraints_);
}
```
then when configuring below:
```cpp
configuration->planar_constraints = planar_constraints_?1:0;
```

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,940 @@
...
+// ============================================================================
+// OdometryCuVSLAM Class Implementation
+// ============================================================================
+
+namespace rtabmap {
+
+OdometryCuVSLAM::OdometryCuVSLAM(const ParametersMap & parameters) :
+    Odometry(parameters),
+#ifdef RTABMAP_CUVSLAM
+    cuvslam_handle_(nullptr),
```
matlabbe:
```suggestion
    Odometry(parameters)
#ifdef RTABMAP_CUVSLAM
    ,cuvslam_handle_(nullptr),
```

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: Still an issue :( Currently looking into fixing this.

corelib/src/odometry/OdometryCuVSLAM.cpp
matlabbe: `configuration->planar_constraints` seems used just for the SLAM part. To constrain odometry, they use another header [ground_constraint.h](https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_nitros/blob/main/isaac_ros_nitros/lib/cuvslam/include/ground_constraint.h). See [isaac visualslam](https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_visual_slam/blob/19be8c781a55dee9cfbe9f097adca3986638feb1/isaac_ros_visual_slam/src/impl/visual_slam_impl.cpp#L60-L86) to know how to use it.

corelib/src/odometry/OdometryCuVSLAM.cpp
matlabbe: resolved by https://github.com/introlab/rtabmap/pull/1583/commits/32d66ab00748c67b7e83d93b44ad036cb9bccaf4

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
```diff
@@ -0,0 +1,83 @@
...
+	//visualization
+	std::vector<CUVSLAM_Observation> observations_;
+	std::vector<CUVSLAM_Landmark> landmarks_;
+	
+	// GPU memory management
+	std::vector<uint8_t *> gpu_left_image_data_; // pointers to all gpu images
+	std::vector<uint8_t *> gpu_right_image_data_;
+	std::vector<size_t> gpu_left_image_sizes_; // size of one image
+	std::vector<size_t> gpu_right_image_sizes_;
+	void * cuda_stream_; // Opaque CUDA stream handle stored as void*
```
matlabbe: Use `cudaStream_t` type directly.

```suggestion
	cudaStream_t cuda_stream_;
```

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1035 @@
...
+namespace rtabmap {
+
+bool initializeCuVSLAM(const SensorData & data, 
+                       CUVSLAM_TrackerHandle & cuvslam_handle,
+                       CUVSLAM_GroundConstraintHandle & ground_constraint_handle,
+                       std::vector<size_t> & gpu_left_image_sizes,
+                       std::vector<size_t> & gpu_right_image_sizes,
+                       std::vector<CUVSLAM_Camera> & cuvslam_cameras,
+	                   std::vector<std::array<float, 12>> & intrinsics,
+                       void * & cuda_stream);
```
matlabbe:
```suggestion
                       cudaStream_t & cuda_stream);
```

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1035 @@
...
+    
+CUVSLAM_Configuration CreateConfiguration(const CUVSLAM_Pose & cv_base_link_pose_cv_imu, const SensorData & data);
+
+bool prepareImages(const SensorData & data, 
+                    std::vector<CUVSLAM_Image> & cuvslam_images,
+                    std::vector<uint8_t *> & gpu_left_image_data,
+                    std::vector<uint8_t *> & gpu_right_image_data,
+                    std::vector<size_t> & gpu_left_image_sizes,
+                    std::vector<size_t> & gpu_right_image_sizes,
+                    void * & cuda_stream);
```
matlabbe:
```suggestion
                    cudaStream_t & cuda_stream);
```

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: added the planar constraints from ground_constraint.h and it set through the odometer parameters.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: Using properly typed cudaStream_t as a member variable.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: Using properly typed cudaStream_t as a member variable.

corelib/include/rtabmap/core/odometry/OdometryCuVSLAM.h
Dekempsy4: Using properly typed cudaStream_t as a member variable.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1016 @@
...
+        cv::cvtColor(left_image, processed_left_image, cv::COLOR_BGR2RGB);
+        left_encoding = CUVSLAM_ImageEncoding::RGB8;
+    } else {
+        UERROR("Unsupported left image format: %d channels", left_image.channels());
+        return false;
+    }
+
+    // Process right image - makes a copy of the image
+    if(right_image.channels() == 1) {
+        processed_right_image = right_image.clone();
```
matlabbe: That clone may not be necessary if the sub images are cloned below

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -0,0 +1,1016 @@
...
+
+    // Convert image format for cuVSLAM - mono8 or rgb8
+    cv::Mat processed_left_image;
+    cv::Mat processed_right_image;
+    CUVSLAM_ImageEncoding left_encoding;
+    CUVSLAM_ImageEncoding right_encoding;
+    
+    // process left image - makes a copy of the image
+    if(left_image.channels() == 1) {
+        processed_left_image = left_image.clone();
```
matlabbe: That clone may not be necessary if the sub images are cloned below

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: I'll test out removing this copying.

Dekempsy4: I removed the cloning. I've implemented a non-copying channel switching to convert from BGR->RGB. I'm not sure how to test it currently.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: same as above

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: I've reverted this change. We decided the new in-place channel switching was needlessly complex. I now make a copy for BGR->RGB conversion but no copy for MONO8 images.

matlabbe:
LGTM!

PR #1603, opened 2025-10-23
SuperPoint 

Dekempsy4:
Implements the pytorch based Superpoint implementation for keypoint and descriptor extraction:
[https://github.com/rpautrat/SuperPoint](https://github.com/rpautrat/SuperPoint). 

I've also updated the standalone application GUI to include a superpoint rpautrat configuration window with instructions on how to generate the model weights manually.
image

### DIFF
--- a/corelib/src/superpoint_rpautrat/superpoint_to_torchscript.py
+++ b/corelib/src/superpoint_rpautrat/superpoint_to_torchscript.py
@@ -95,47 +95,6 @@ if __name__ == "__main__":
     parser.add_argument("--threshold", type=float, default=0.005, help="Confidence threshold")
     args = parser.parse_args()
     print(f"Generating model from weights: {args.weights} to output: {args.output}")
-
-
-    
-    # INSERT_YOUR_CODE
-    print("============================================================")
-    print(" superpoint_to_torchscript.py - MAIN ENTRYPOINT")
-    print("============================================================")
-    print(" Arguments received for model generation:")
-    print("  --weights:    ", args.weights)
-    print("  --output:     ", args.output)
-    print("  --cuda:       ", args.cuda)
-    print("  --width:      ", args.width)
-    print("  --height:     ", args.height)
-    print("  --nms_radius: ", args.nms_radius)
-    print("  --threshold:  ", args.threshold)
-    print("------------------------------------------------------------")
-    print("   Did you mean to run this as a standalone script?")
-    print("   If you see this, then the script is being used!")
-    print("   (This block intentionally verbose for debugging.)")
-    print("------------------------------------------------------------")
-    print(" Python version:", __import__("sys").version)
-    try:
-        import torch
-        print(" Torch version:", torch.__version__)
-        print("   CUDA available:", torch.cuda.is_available())
-        print("   CUDA devices:  ", torch.cuda.device_count() if torch.cuda.is_available() else 0)
-    except Exception as e:
-        print(" ERROR: Unable to import torch: ", e)
-    try:
-        import os
-        print(" Current working dir:", os.getcwd())
-    except Exception as e:
-        print(" ERROR: Unable to import os: ", e)
-    print("------------------------------------------------------------")
-    print(" WARNING: This script may be called programmatically by RTAB-Map!")
-    print("   If you see this output in your logs, model export should work.")
-    print("============================================================")
-    import time
-    time.sleep(0.5)
-
-
     
     generate_model(
         weights_path=args.weights,


corelib/include/rtabmap/core/Features2d.h
```diff
@@ -129,7 +130,8 @@ class RTABMAP_CORE_EXPORT Feature2D {
 		kFeatureSurfFreak=12, //new 0.20.4
 		kFeatureGfttDaisy=13, //new 0.20.6
 		kFeatureSurfDaisy=14,  //new 0.20.6
-		kFeaturePyDetector=15}; //new 0.20.8
+		kFeaturePyDetector=15, //new 0.20.8
+		kFeatureSuperPointRpautrat=16};
```
matlabbe: Bump the patch version here https://github.com/introlab/rtabmap/blob/a3076392b5fc4273a6d42f03261bfc6281c02a58/CMakeLists.txt#L23 and add comment with that version here like the other ones.

corelib/src/python/rtabmap_superpoint.py
matlabbe, line 1: These changes seem unrelated to the new approach

corelib/src/python/rtabmap_superpoint_rpautrat.py
```diff
@@ -0,0 +1,73 @@
+#! /usr/bin/env python3
+#
+# Drop this file in the root folder of SuperPoint git: https://github.com/rpautrat/SuperPoint
+# To use with rtabmap:
+#   --Vis/FeatureType 15 --Kp/DetectorStrategy 15 --PyDetector/Path "~/SuperPoint/rtabmap_superpoint_rpautrat.py"
```
matlabbe, line 5:
```suggestion
#   --Vis/FeatureType 16 --Kp/DetectorStrategy 16 --PyDetector/Path "~/SuperPoint/rtabmap_superpoint_rpautrat.py"
```

corelib/src/python/rtabmap_superpoint_rpautrat.py
```diff
@@ -0,0 +1,73 @@
...
+    # Ensure model is in eval mode for inference
+    superpoint.eval()
+
+def detect(imageBuffer):
+    global superpoint, device
+    
+    image = np.asarray(imageBuffer)
+    image = (image.astype('float32') / 255.)
+    
+    # image_tensor = torch.from_numpy(image[None, None]).float().to(device)
```
matlabbe: cleanup comments

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
```diff
@@ -0,0 +1,247 @@
...
+static std::string exportSuperPointTorchScript(
+    const std::string & superpointDir,
+    const int & width,
+    const int & height,
+    const float & threshold,
+    const int & nms_radius,
+    const bool & cuda)
+{
+    // Resolve paths
+    const std::string srcDir = UDirectory::getDir(__FILE__); // folder of this .cpp at build time
```
matlabbe: That won't work when rtabmap is installed and/or deployed. I think the best would be to include `superpoint_to_torchscript.py` as a "resource" in rtabmap library. That can be done by appending `superpoint_to_torchscript.py` to RESOURCES here https://github.com/introlab/rtabmap/blob/a3076392b5fc4273a6d42f03261bfc6281c02a58/corelib/src/CMakeLists.txt#L829 (only append if we are building with Torch and python).

The way it works afterwards is that you can include the resource file and decode it into a string:
```cpp
#include "superpoint_to_torchscript_py.h"
[...]
std::string file_content = uHex2Str(SUPERPOINT_TO_TORSCRIPT_PY);
```

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
```diff
@@ -0,0 +1,247 @@
...
+        << " --width " << width
+        << " --height " << height
+        << " --weights \"" << weights << "\""
+        << " --output \"" << output << "\""
+        << " --threshold " << threshold
+        << " --nms_radius " << nms_radius << " "
+        << cuda_flag;
+        
+    UINFO("Executing: %s", cmd.str().c_str());
+    const int code = std::system(cmd.str().c_str());
```
matlabbe: Not sure if the windows CI will pass with this, and I would try to stay away from system calls for security reason. Could pybind be used to call the conversion script?

corelib/src/CMakeLists.txt
```diff
@@ -214,10 +214,12 @@ IF(TORCH_FOUND)
 	SET(SRC_FILES
 		${SRC_FILES}
 		superpoint_torch/SuperPoint.cc
+		superpoint_rpautrat/SuperpointRpautrat.cpp
```
matlabbe: SuperpointRpautrat may be added also only if we build with python too.

corelib/include/rtabmap/core/Parameters.h
```diff
@@ -346,6 +346,12 @@ class RTABMAP_CORE_EXPORT Parameters
     RTABMAP_PARAM(SuperPoint, NMSRadius,     int,  4,      uFormat("[%s=true] Minimum distance (pixels) between keypoints.", kSuperPointNMS().c_str()));
     RTABMAP_PARAM(SuperPoint, Cuda,          bool, true,   "Use Cuda device for Torch, otherwise CPU device is used by default.");
 
+    RTABMAP_PARAM_STR(SuperPointRpautrat, Dir, "",           "[Required] SuperPoint root directory containing scripts and weights.");
```
matlabbe: I would suggest to use a parameter for the `pth` file directly: `SuperPointRpautrat/modelPath`, instead of a directory. See my other comment about which directory to use to output the converted weights.

corelib/src/Features2d.cpp
```diff
@@ -2615,6 +2619,91 @@ cv::Mat SuperPointTorch::generateDescriptorsImpl(const cv::Mat & image, std::vec
...
+void SuperPointRpautrat::parseParameters(const ParametersMap & parameters)
+{
+	Feature2D::parseParameters(parameters);
+
+	std::string previousPath = superpointDir_;
+#ifdef RTABMAP_TORCH
+	bool previousCuda = cuda_;
+#endif
+	
+	Parameters::parse(parameters, Parameters::kSuperPointRpautratDir(), superpointDir_);
```
matlabbe: The output file could be generated in the RTABMap's working directory: available by parsing `Rtabmap/WorkingDirectory` parameter.
```cpp
superpointDir_ = Parameters::parse(parameters, Parameters::kRtabmapWorkingDirectory(), "");
```
As suggested above, you may have also to add:
```
Parameters::parse(parameters, Parameters::kSuperPointRpautratModelPath(), superpointModelPath_);
```

corelib/src/Features2d.cpp
```diff
@@ -2615,6 +2619,91 @@ cv::Mat SuperPointTorch::generateDescriptorsImpl(const cv::Mat & image, std::vec
...
+	{
+		superPoint_ = cv::Ptr<SPDetectorRpautrat>(new SPDetectorRpautrat(superpointDir_, threshold_, nms_, minDistance_, cuda_));
+	}
+	else
+	{
+		superPoint_->setSuperpointDir(superpointDir_);
+		superPoint_->setThreshold(threshold_);
+		superPoint_->setNMS(nms_);
+		superPoint_->setMinDistance(minDistance_);
+	}
```
matlabbe: You may remove these setters because they are not really supported. The model is initialized on the first frame, so any changes here won't matter, the whole model should be re-generated again to use of the new parameters.

It seems that if any parameter changes, we need to delete the previous detector to force it re-initialize on the next frame.

guilib/src/ui/preferencesDialog.ui
```diff
@@ -27861,6 +27871,166 @@ Lower the ratio -&gt; higher the precision.</string>
...
+                      <property name="text">
+                       <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
+&lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
+p, li { white-space: pre-wrap; }
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Ubuntu'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;SuperPoint C++ implementation from the &lt;a href=&quot;https://github.com/rpautrat/SuperPoint&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#0000ff;&quot;&gt;rpautrat/SuperPoint&lt;/span&gt;&lt;/a&gt; project.&lt;/p&gt;
+&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;A superpoint_v6_cpp.pt model file will be generated automatically by using the SuperPointRpautrat detector as long as RTAB-Map is built with python support.&lt;/p&gt;
```
matlabbe:
```suggestion
&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;A "pt" model file with same name than the "pth" file will be generated automatically in RTAB-Map's working directory so that the model can be called by c++ libtorch directly.&lt;/p&gt;
```

guilib/src/ui/preferencesDialog.ui
```diff
@@ -27861,6 +27871,166 @@ Lower the ratio -&gt; higher the precision.</string>
...
+&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;To generate the TorchScript model file manually, clone the repository at &lt;a href=&quot;https://github.com/rpautrat/SuperPoint&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#0000ff;&quot;&gt;https://github.com/rpautrat/SuperPoint.&lt;/span&gt;&lt;/a&gt; Copy the script from &lt;span style=&quot; font-family:'monospace'; background-color:#ffff99;&quot;&gt;corelib/src/superpoint_rpautrat/superpoint_to_torchscript.py&lt;/span&gt; into the SuperPoint repo. Then run the script with the desired parameters, e.g.&lt;br /&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;python superpoint_to_torchscript.py --width WIDTH --height HEIGHT --weights weights/superpoint_v6_from_tf.pth --output weights/superpoint_v6_cpp.pt&lt;/span&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
matlabbe: I would remove this. We won't support the manual way.

guilib/src/PreferencesDialog.cpp
```diff
@@ -290,6 +290,8 @@ PreferencesDialog::PreferencesDialog(QWidget * parent) :
 #ifndef RTABMAP_TORCH
 	_ui->comboBox_detector_strategy->setItemData(11, 0, Qt::UserRole - 1);
 	_ui->vis_feature_detector->setItemData(11, 0, Qt::UserRole - 1);
+	_ui->comboBox_detector_strategy->setItemData(16, 0, Qt::UserRole - 1);
+	_ui->vis_feature_detector->setItemData(16, 0, Qt::UserRole - 1);
```
matlabbe, line 299: #if !defined(RTABMAP_TORCH) || !defined(RTABMAP_PYTHON)

guilib/src/ui/preferencesDialog.ui
matlabbe: Also: "RTAB-Map should be built with libtorch and python to use this feature."

matlabbe:
In the top CMakeLists.txt around here https://github.com/introlab/rtabmap/blob/a3076392b5fc4273a6d42f03261bfc6281c02a58/CMakeLists.txt#L1455-L1461

Can you add a section for SuperPoint Rpautrat? Maybe like:
```cmake
IF(TORCH_FOUND AND WITH_PYTHON AND Python3_FOUND) 
 MESSAGE(STATUS "  With SuperPoint Rpautrat = YES (License: MIT)") 
 ELSEIF(NOT WITH_TORCH) 
 MESSAGE(STATUS "  With SuperPoint Rpautrat = NO (WITH_TORCH=OFF)") 
 ELSEIF(NOT WITH_PYTHON) 
 MESSAGE(STATUS "  With SuperPoint Rpautrat = NO (WITH_PYTHON=OFF)") 
 ELSE() 
 MESSAGE(STATUS "  With SuperPoint Rpautrat = NO (libtorch and/or python3 not found)") 
 ENDIF() 
```

matlabbe:
Similarly to previous comment: the about dialog could also be updated to show if rtabmap is built with superpoint rpautrat
image

For Parameters.cpp, we can also add the status to know if rtabmap is built with tht feature: https://github.com/introlab/rtabmap/blob/a3076392b5fc4273a6d42f03261bfc6281c02a58/corelib/src/Parameters.cpp#L659-L664
```cpp
 				str = "With SuperPoint Rpautrat:"; 
 #if defined(RTABMAP_TORCH) and defined(RTABMAP_PYTHON) 
 				std::cout << str << std::setw(spacing - str.size()) << "true" << std::endl; 
 #else 
 				std::cout << str << std::setw(spacing - str.size()) << "false" << std::endl; 
 #endif

```

corelib/include/rtabmap/core/Features2d.h
Dekempsy4: Bumped the patch and added a comment, (0.23.3)

corelib/src/python/rtabmap_superpoint.py
Dekempsy4: I found these changes are necessary to get the old SuperPoint to work. They are unrelated to the focus of this MR so I can remove them if preferred.

corelib/src/python/rtabmap_superpoint_rpautrat.py
Dekempsy4: This is the python impl which relies on the detector strategy to be 15 (Pydetector), so I believe that this is correct.

corelib/src/python/rtabmap_superpoint_rpautrat.py
Dekempsy4: Whoops, removed the comment.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4: Updated to load the file as a resource. I'm generating the loaded .py file in the rtabmap working dir, but it is cleaned up as soon as the model file is generated.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4: I'm now executing using pybind11 through runpy.
```cpp
auto runpy = pybind11::module_::import("runpy");
runpy.attr("run_path")(dstScript, pybind11::arg("run_name") = "__main__");
```

corelib/src/CMakeLists.txt
Dekempsy4: Now only add superpoint rpautrat if we have python and torch.

corelib/include/rtabmap/core/Parameters.h
Dekempsy4: I've changed it so that we expect the path to the .pth weights file instead of the SuperPoint repo. I originally expected the SuperPoint dir because we need to set the PYTHONPATH there in order for the ```superpoint_pytorch.py``` python module import to resolve.

Currently, the code still makes assumptions about the structure of the SuperPoint directory in order to set the PYTHONPATH correctly.
```cpp
// Add both the weights directory and its parent (repo root) to sys.path
// This assumes a certain direcotry structure: /SuperPoint/weights/superpoint_v6_from_tf.pth
std::string weightsDir = UDirectory::getDir(weightsPath);
std::string repoRoot = UDirectory::getDir(weightsDir);
auto sys = pybind11::module_::import("sys");
auto sysPath = sys.attr("path").cast<pybind11::list>();
sysPath.append(repoRoot);
sysPath.append(weightsDir);
``` 

The most robust solution would be to expect a path to the SuperPoint model class AND the weights file as params. I can implement this if requested.

corelib/src/Features2d.cpp
Dekempsy4: I'm now forwarding the kRtabmapWorkingDirectory parameter to Features2D and generating the model file there.

corelib/src/Features2d.cpp
Dekempsy4: I've removed the setters, we will now regenerate the detector when any parameter changes.

guilib/src/ui/preferencesDialog.ui
Dekempsy4: I updated the info text in the GUI. The MR comment above is updated as well.

guilib/src/ui/preferencesDialog.ui
Dekempsy4: Removed!

guilib/src/PreferencesDialog.cpp
Dekempsy4: Now checking if RTAB-Map is built with Python and Torch everywhere for SuperPoint rpautrat to be used.

Dekempsy4:
@matlabbe I cannot add comments on the last two threads for some reason, but I addressed both of them. This MR should be ready for another round of review when you have time.

corelib/include/rtabmap/core/Parameters.h
matlabbe: Oh I see, but I wouldn't hard-code `superpoint_v6_from_tf.pth`, if different weights file are generated for some reason. Add a parameter for the [superpoint_pytorch.py](https://github.com/rpautrat/SuperPoint/blob/master/superpoint_pytorch.py) script, then derive its directory like you do above.

Is `weightsDir` required to be added to the path? `torch.load` may just need the filepath.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
```diff
@@ -0,0 +1,287 @@
...
+			UERROR("Failed to open destination script for writing: %s", dstScript.c_str());
+			return "";
+		}
+		ofs << file_content;
+		if(!ofs.good())
+		{
+			UERROR("Failed to write embedded script to: %s", dstScript.c_str());
+			return "";
+		}
+	}
```
matlabbe: could it be possible to avoid making a file so that pybind re-open it afterwards to get its content back? It seems that pybind has some utilities to execute a script from a string: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html#evaluating-python-expressions-from-strings-and-files
```cpp
py::exec(
    uHex2Str(SUPERPOINT_TO_TORCHSCRIPT_PY),
    scope);
```

matlabbe: Based on gemini answer, we can pass arguments like this:
```
py::dict globals = py::globals(); // Get the current global scope
// Or create a new dictionary for a more isolated scope:
// py::dict locals; 
    
globals["my_argument"] = 42; // Pass an integer argument
globals["another_string"] = "hello"; // Pass a string argument

py::exec("print(f'Hello from string! Value: {my_argument}')", globals);
```

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
matlabbe: Follow up in that comment: https://github.com/introlab/rtabmap/pull/1603/files#r2487984815

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
matlabbe: follow-up comment: https://github.com/introlab/rtabmap/pull/1603/files#r2487984815

corelib/src/python/rtabmap_superpoint.py
matlabbe: That example https://github.com/introlab/rtabmap_ros/tree/ros2/docker/jazzy/superpoint works, so not sure if it is computer config specific related. I would remove that change from this MR.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
```diff
@@ -0,0 +1,287 @@
...
+            float dy = y - stored_y;
+            float distSq = dx * dx + dy * dy;
+            
+            if(distSq < 1.0f) {
+                filtered_descriptors[i] = desc_[j];
+                break;
+            }
+        }
+    }
+
```
matlabbe: I would remove this and put an assert that the keypoints query size should be the same size than the current descriptors, similar to https://github.com/introlab/rtabmap/blob/f44a4fc4786bd3c859a599718f42d3dee476ba8a/corelib/src/python/PyDetector.cpp#L226

then modify https://github.com/introlab/rtabmap/blob/f44a4fc4786bd3c859a599718f42d3dee476ba8a/corelib/src/Features2d.cpp#L834:
```
if (this->getType() != Feature2D::Type::kFeaturePyDetector && this->getType() != Feature2D::Type::kFeatureSuperpointRpautrat )
```
just to make sure, though it should the number of features should be already limited by `Kp/MaxFeatures` passed to superpoint initialization already.

corelib/src/Features2d.cpp
```diff
@@ -2615,6 +2623,100 @@ cv::Mat SuperPointTorch::generateDescriptorsImpl(const cv::Mat & image, std::vec
...
+SuperPointRpautrat::~SuperPointRpautrat()
+{
+}
+
+void SuperPointRpautrat::parseParameters(const ParametersMap & parameters)
+{
+	Feature2D::parseParameters(parameters);
+
+	std::string previousPath = superpointWeightsPath_;
+#if defined(RTABMAP_TORCH) && defined(RTABMAP_PYTHON)
```
matlabbe, line 2649: nit: you could enclose the whole function implementation under the same preprocessor if.

corelib/src/CMakeLists.txt
```diff
@@ -836,6 +846,7 @@ SET(RESOURCES
 	${CMAKE_CURRENT_SOURCE_DIR}/resources/backward_compatibility/DatabaseSchema_0_16_2.sql
 	${CMAKE_CURRENT_SOURCE_DIR}/resources/backward_compatibility/DatabaseSchema_0_16_1.sql
 	${CMAKE_CURRENT_SOURCE_DIR}/resources/backward_compatibility/DatabaseSchema_0_16_0.sql
+	${CMAKE_CURRENT_SOURCE_DIR}/superpoint_rpautrat/superpoint_to_torchscript.py
 )
```
matlabbe:
```suggestion
	${CMAKE_CURRENT_SOURCE_DIR}/superpoint_rpautrat/superpoint_to_torchscript.py
)
IF(TORCH_FOUND AND WITH_PYTHON AND Python3_FOUND)
	SET(RESOURCES 
		${RESOURCES}
		${CMAKE_CURRENT_SOURCE_DIR}/superpoint_rpautrat/superpoint_to_torchscript.py
	)
ENDIF()
```

matlabbe:
The AboutDialog part of this comment https://github.com/introlab/rtabmap/pull/1603#issuecomment-3469747044 seems not done yet

corelib/src/python/rtabmap_superpoint.py
Dekempsy4: Okay understood, I removed these changes.

corelib/include/rtabmap/core/Parameters.h
Dekempsy4: I've added another parameter for the superpoint_pytorch.py model file so that we make no assumptions about the the layout of the SuperPoint repository.

Dekempsy4: I'm not adding the weights dir to sys.path anymore also, this wasn't necessary.l

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4: I've updated this section to execute the script from a string instead of writing to a file.

Apparently the global scope already has access to the sys.argv so we can parse the commands in the same way. 

```cpp
// Build sys.argv for the script
pybind11::list argv;
argv.append("superpoint_to_torchscript.py");  // Use the model filename as script name
argv.append("--width");       
argv.append(std::to_string(width));
argv.append("--height");      
argv.append(std::to_string(height));
argv.append("--weights");     
argv.append(weightsPath);
argv.append("--output");      
argv.append(output);
argv.append("--threshold");   
argv.append(std::to_string(threshold));
argv.append("--nms_radius");  
argv.append(std::to_string(nms_radius));
if(cuda) { 
    argv.append("--cuda"); 
}
sys.attr("argv") = argv;

// Run the script
auto main_module = pybind11::module_::import("__main__");
main_module.attr("__name__") = "__main__";

pybind11::exec(uHex2Str(SUPERPOINT_TO_TORCHSCRIPT_PY));
pybind11::exec("main()");
```

corelib/src/Features2d.cpp
Dekempsy4: Now using only preprocessor if in this block.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4: As we discussed, I removed the spatial matching logic but we will not use the SuperPoint top-K logic because certain pathways cannot be captured in the model trace. This leads to crashes when the detector returns less keypoints than the maximum feature param. Instead we will rely on top-K sampling from upstream logic and return the full keypoints and descriptors from the detector.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
matlabbe: ~Where is `sys.attr("argv")` saved? would these arguments passed to any other similar python calls?~

I see in the code:
```cpp
auto sys = pybind11::module_::import("sys");
auto sysPath = sys.attr("path").cast<pybind11::list>();
sysPath.append(modelDir);
```

however, I still wonder how `pybind11::exec` knows about sys, and if any other `pybind11::exec` would see the same thing? even after `sys` is destroyed...

I think if `sysPath.append(modelDir);` is set globally, that could be an issue if we change the model directory, so both model directories would be added and probably have two SuperPoint imported. If `sys.argv` is set for all subsequent `pybind11::exec()` calls, you may pass the arguments as "globals", or probably more 'locals" to give less scope. Instead of calling `main`, you may create a function in python script like:
```py
export_torchscript(weigths, output, cuda, widt, height, nms_radius, threshold):
    #code...

main():
    parser = argparse.ArgumentParser(description="Convert SuperPoint weights to TorchScript")
    parser.add_argument("--weights", required=True, help="Path to weights file")
    parser.add_argument("--output", required=True, help="Output TorchScript file")
    parser.add_argument("--cuda", action="store_true", help="Use CUDA")
    parser.add_argument("--width", type=int, default=1920, help="Width of the input image")
    parser.add_argument("--height", type=int, default=288, help="Height of the input image")
    parser.add_argument("--nms_radius", type=int, default=4, help="NMS radius")
    parser.add_argument("--threshold", type=float, default=0.005, help="Confidence threshold")
    args = parser.parse_args()
    export_torchscript(args.weights, args.output, args.cuda, args.width, args.height, args.nms_radius, args.threshold)
```

then call `export_torchscript` from c++.
```cpp
   pybind11::dict scope;
    scope["__builtins__"] = pybind11::module::import("builtins");
    try {
        pybind11::exec(uHex2Str(SUPERPOINT_TO_TORCHSCRIPT_PY), scope, scope);

        // Retrieve the function
        pybind11::function export_torchscript_func = scope["export_torchscript"].cast<pybind11::function>();

        // Call the function with arguments 
        pybind11::object result = export_torchscript_func(weights, output, cuda, width, height, nms_radius, threshold);
    } catch (const pybind11::error_already_set &e) {
        UERROR("Python error: %s", e.what());
    }
```

It seems we can temporary set the sys path for a function call. 
```cpp
std::string custom_path = "/path/to/your/python/modules"; 
// Get the 'sys.path' list
pybind11::module sys = pybind11::module::import("sys");
pybind11::list sys_path = sys.attr("path");
        
//  Append the custom path to the global sys.path (temporarily for this session)
sys_path.attr("insert")(0, custom_path);

// Call python function
.... pybind11::exec() ....

// Remove the path from sys.path to restore the original state
// In a complex application, Python's "context manager" approach is safer, 
// but simple removal works for this use case.
sys_path.attr("remove")(custom_path); // Use remove() if inserted, pop(0) if prepended
```

*Some code generated by gemini...

Dekempsy4: Okay I'm now using a local dict to isolate the execution so it won't affect any other python modules or any regeneration of the model file. We also remove the SuperPoint dir from the sys.path after running the script.

corelib/src/CMakeLists.txt
Dekempsy4: I updated this CMake to only add the superpoint_to_torchscript.py file to resources if we are building with superpoint rpautrat support.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4:
```cpp

    // Execute the script inside the embedded Python interpreter
    try
    {
        pybind11::gil_scoped_acquire acquire;
        pybind11::dict scope;
        scope["__builtins__"] = pybind11::module_::import("builtins");
        
        // set sys.path to the location of the model definition so it can be imported
        std::string model_dir = UDirectory::getDir(modelPath);
        auto sys = pybind11::module_::import("sys");
        pybind11::list sys_path = sys.attr("path");
        sys_path.attr("insert")(0, model_dir);
        
        try {
            // execute the script to generate the model
            pybind11::exec(uHex2Str(SUPERPOINT_TO_TORCHSCRIPT_PY), scope, scope);
            pybind11::function generate_model = scope["generate_model"].cast<pybind11::function>();
            pybind11::object result = generate_model(weightsPath, output, cuda, nms_radius, threshold, width, height);
            sys_path.attr("remove")(model_dir);
        }
        catch(...) {
            // Ensure sys.path cleanup on any exception
            sys_path.attr("remove")(model_dir);
            throw;
        }

    }
    // pybind11 throws std::exception for RuntimeError
    catch (const std::exception &e)
    {
        UERROR("Python export failed: %s", e.what());
        return "";
    }
```

Dekempsy4:
@matlabbe I believe I've addressed all the 2nd round review comments now. Let me know if there are any changes you would like!

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
matlabbe: LGTM!

matlabbe:
Great thank you!

PR #1614, opened 2025-11-13
Limit Max Features

Dekempsy4:
This MR is a small improvement to the PyDetector and SuperPointRpautrat detector strategies.

We now apply the Feature2D::limitKeypoints method inside the detectors themselves rather then deferring this to an upstream process. This reduces the size of the detector and key-point data-structures and should marginally improve performance.

### DIFF
--- a/corelib/src/python/PyDetector.cpp
+++ b/corelib/src/python/PyDetector.cpp
@@ -156,12 +156,12 @@ std::vector<cv::KeyPoint> PyDetector::generateKeypointsImpl(const cv::Mat & imag
 	if(pFunc_)
 	{
 		npy_intp dims[2] = {imgRoi.rows, imgRoi.cols};
-		PyObject* pImageBuffer = PyArray_SimpleNewFromData(2, dims, NPY_UBYTE, (void*)imgRoi.data);
+		PyObject * pImageBuffer = PyArray_SimpleNewFromData(2, dims, NPY_UBYTE, (void*)imgRoi.data);
 		UASSERT(pImageBuffer);
 
 		UDEBUG("Preparing data time = %fs", timer.ticks());
 
-		PyObject *pReturn = PyObject_CallFunctionObjArgs(pFunc_, pImageBuffer, NULL);
+		PyObject * pReturn = PyObject_CallFunctionObjArgs(pFunc_, pImageBuffer, NULL);
 		if(pReturn == NULL)
 		{
 			UERROR("Failed to call match() function!");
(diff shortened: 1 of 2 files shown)


corelib/src/Features2d.cpp
```diff
@@ -2653,6 +2655,8 @@ void SuperPointRpautrat::parseParameters(const ParametersMap & parameters)
 	float previousThreshold = threshold_;
 	bool previousNms = nms_;
 	int previousMinDistance = minDistance_;
+	int previousMaxFeatures = maxFeatures_;
+	bool previousSsc = ssc_;
```
matlabbe: You may not have to create two new variables `maxFeatures_` and `ssc_`. You could move this above `Feature2D::parseParameters(parameters);`:

```
int previousMaxFeatures = this->getMaxFeatures();
bool previousSsc =  this->getSSC();

Feature2D::parseParameters(parameters);

...

if(outputDir_.empty())
	{
	   previousCuda != cuda_ ||
	   previousThreshold != threshold_ ||
	   previousNms != nms_ ||
	   previousMinDistance != minDistance_ ||
	   previousMaxFeatures != this->getMaxFeatures() ||
	   previousSsc != this->getSSC())
	{
		superPoint_ = cv::Ptr<SPDetectorRpautrat>(new SPDetectorRpautrat(superpointWeightsPath_, 
            superpointModelPath_, 
            outputDir_, 
             threshold_, 
             nms_, 
             minDistance_, 
             cuda_, 
             this->getMaxFeatures(), 
             this->getSSC()));

```

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
```diff
@@ -136,12 +139,10 @@ cv::Mat SPDetectorRpautrat::compute(const std::vector<cv::KeyPoint> &keypoints)
 	}
 
     // These should have the same size
-    UASSERT(static_cast<size_t>(desc_.size(0)) == keypoints.size());
+    UASSERT(static_cast<size_t>(desc_.rows) == keypoints.size());
 
     // Move to CPU and return descriptors computed in the forward pass
```
matlabbe: That comment is probably not needed anymore

corelib/src/Features2d.cpp
Dekempsy4: There's an issue with this setup related to [this section in memory](https://github.com/introlab/rtabmap/blob/c919f116d7cc88eeef05b44b57cdf2d8effc22fb/corelib/src/Memory.cpp#L5105C5-L5133C63). When we don't set Kp/MaxFeatures to anything this section may change the param, which triggers a reset to the detector. Then it attempts to call `compute()`which fails since the detector was reset since the last call to `detect()`.

To fix this I made it so that `maxFeatures_` and `ssc_` can be reconfigured through setters without triggering a full reset.

corelib/src/superpoint_rpautrat/SuperpointRpautrat.cpp
Dekempsy4: I removed the leftover comment

corelib/src/Features2d.cpp
matlabbe: Sounds good

matlabbe (approved):

PR #1618, opened 2025-11-18
CuVSLAM 14 Update

Dekempsy4:
Updating CuVSLAM implementation to match the new v4.0 isaac_ros_nitros release:
[https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_nitros/tree/v4.0-0](https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_nitros/tree/v4.0-0)

Also added a CMAKE error when the cuVSLAM major version doesn't match the expected.


cmake_modules/FindCuVSLAM.cmake
```diff
@@ -4,6 +4,7 @@
 #
 # It sets the following variables:
 #  CUVSLAM_FOUND         - Set to false, or undefined, if cuVSLAM isn't found.
+#  CuVSLAM_VERSION       - The major version of cuVSLAM found (e.g., "14").
```
matlabbe: We should expect format like `14.0.0`

The minor version seems available: https://github.com/NVIDIA-ISAAC-ROS/isaac_ros_nitros/blob/8d2a14f301f8b169626637c8d87feb8317e22369/isaac_ros_nitros/lib/cuvslam/include/cuvslam.h#L40-L44

keep 0 for the patch.

cmake_modules/FindCuVSLAM.cmake
```diff
@@ -46,11 +57,23 @@ if(CUVSLAM_INCLUDE_DIRS AND CUVSLAM_LIBRARY)
...
+# Version checking - enforce exact version match for major version
+if(CUVSLAM_FOUND AND CuVSLAM_FIND_VERSION AND CUVSLAM_VERSION)
+    if(NOT CUVSLAM_VERSION VERSION_EQUAL CuVSLAM_FIND_VERSION)
+        message(FATAL_ERROR 
+            "cuVSLAM version mismatch: found version ${CUVSLAM_VERSION} but version ${CuVSLAM_FIND_VERSION} is required.\n"
+            "cuVSLAM does not guarantee backward compatibility between major versions.\n"
+            "Please install cuVSLAM version ${CuVSLAM_FIND_VERSION} or update CMakeLists.txt to request version ${CUVSLAM_VERSION}."
+        )
+    endif()
+endif()
```
matlabbe: Doublecheck, but [find_package_handle_standard_args](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) should already do this check with `VERSION_VAR <version-var>`

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -726,19 +781,15 @@ CUVSLAM_Configuration CreateConfiguration(const SensorData & data)
...
-    configuration.debug_imu_mode = 0;                   // Disable IMU debug mode
-    // imu_calibration.gyroscope_noise_density = 0.0002f;    
-    // imu_calibration.gyroscope_random_walk = 0.00003f;      
-    // imu_calibration.accelerometer_noise_density = 0.01f;   
-    // imu_calibration.accelerometer_random_walk = 0.001f;    
-    // imu_calibration.frequency = 200.0f; 
-    // configuration.imu_calibration = imu_calibration;
+    // Odometry configuration (Vision-only, no IMU)
+    configuration.odometry_mode = CUVSLAM_OdometryMode::Multicamera;
+    configuration.multicam_mode = 1;
```
matlabbe: Can the code still support single stereo camera?

cmake_modules/FindCuVSLAM.cmake
Dekempsy4: It seems the default behavior of find_package_handle_standard_args only throws an error when the version received is older then the version requested. This would allow a major version that is greater then the requested version.
So if we request version 13.x but receive version 14.x it will pass, but cuvslam doesn't guarantee that major versions are compatible.

Therefore, I added some custom logic so that we still get a helpful failure message in these cases, but rely on the find_package_handle_standard_args when possible. 

```bash
Could NOT find CuVSLAM (missing: CUVSLAM_LIBRARIES) (found suitable version "14.0.0", minimum required is "13.0.0")
Reason given by package: Major version mismatch: found 14.x but requested 13.x.
cuVSLAM only guarantees API compatibility within the same major version.
Please install cuVSLAM 13.x or update CMakeLists.txt to request version 14.0.0
```

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: Yes it can. The CUVSLAM_OdometryMode::Multicamera just indicates that we are using stereo data with no IMU. This supports 1 or more stereo pairs.

cmake_modules/FindCuVSLAM.cmake
matlabbe: I think for this case, one should use `EXACT` keyword or versionMin... versionMax approach when calling [find_package](https://cmake.org/cmake/help/latest/command/find_package.html). I would let the standard cmake handles this.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -247,8 +243,21 @@ OdometryCuVSLAM::~OdometryCuVSLAM()
 
 void OdometryCuVSLAM::reset(const Transform & initialPose)
 {
+    UWARN("=== OdometryCuVSLAM::reset() CALLED === initialPose: %s", initialPose.prettyPrint().c_str());
+    UWARN("RESET: Before - lost_=%s, tracking_=%s, initialized_=%s",
+          lost_ ? "true" : "false",
+          tracking_ ? "true" : "false",
+          initialized_ ? "true" : "false");
+
```
matlabbe: Is it still necessary? or put a UDEBUG instead.

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -434,51 +457,41 @@ Transform OdometryCuVSLAM::computeTransform(
...
+        if(diag_val > 0.1) {
+            valid_covariance = false;
+
+            // If we don't have a guess, we can't use velocity difference to detect lost state.
+            // Thus at this point, we are lost. Warn the user that cuVSLAM probably needs a guess to work well.
+            if(guess.isNull()) {
+                UERROR("No guess provided, but covariance is invalid: %.8f", diag_val);
+                UERROR("We cannot use velocity difference to detect lost state without a guess!");
+                UERROR("Without a guess cuVSLAM is prone to getting lost easily!");
+                UERROR("It is highly recommended to provide a guess to cuVSLAM!");
```
matlabbe: Maybe change this to UWARN

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -726,19 +807,15 @@ CUVSLAM_Configuration CreateConfiguration(const SensorData & data)
...
-    configuration.debug_imu_mode = 0;                   // Disable IMU debug mode
-    // imu_calibration.gyroscope_noise_density = 0.0002f;    
-    // imu_calibration.gyroscope_random_walk = 0.00003f;      
-    // imu_calibration.accelerometer_noise_density = 0.01f;   
-    // imu_calibration.accelerometer_random_walk = 0.001f;    
-    // imu_calibration.frequency = 200.0f; 
-    // configuration.imu_calibration = imu_calibration;
+    // Odometry configuration (Vision-only, no IMU)
+    configuration.odometry_mode = CUVSLAM_OdometryMode::Multicamera;
+    configuration.multicam_mode = 0;    // moderate (0), performance (1) or precision (2).
```
matlabbe: Would it worth it to add this as a RTAB-Map's parameter? If you tried them, did you see any difference?

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -726,19 +807,15 @@ CUVSLAM_Configuration CreateConfiguration(const SensorData & data)
...
+    configuration.debug_imu_mode = 0;
     
-    // configuration.max_frame_delta_ms = 100.0;        // Maximum frame interval (100ms default)
+    // Frame timing
+    configuration.max_frame_delta_s = 0.2f;             // 200ms max frame interval (5Hz minimum)
```
matlabbe: Is this config required? I would remove this constraint if possible.

matlabbe:
It seems the build without CUVSLAM is failing, see CI erorr logs

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: These comments are from my debugging session, I removed them.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: Changed to UWARN

corelib/src/odometry/OdometryCuVSLAM.cpp
```diff
@@ -204,7 +192,14 @@ OdometryCuVSLAM::OdometryCuVSLAM(const ParametersMap & parameters) :
 {
 #ifdef RTABMAP_CUVSLAM
     Parameters::parse(parameters, Parameters::kRegForce3DoF(), planar_constraints_);
-
+	Parameters::parse(parameters, Parameters::kOdomCuVSLAMMulticamMode(), multicam_mode_);
+	if(multicam_mode_ < 0 || multicam_mode_ > 2)
+	{
+		UWARN("%s=%d is invalid, clamping to [0,2].",
+				Parameters::kOdomCuVSLAMMulticamMode().c_str(), multicam_mode_);
+		multicam_mode_ = multicam_mode_ <= 0 ? 0 : 2;
```
matlabbe: Is `1` possible?

You may replace this "if" section by an assert:
```
UASSERT(multicam_mode_ >= 0 && multicam_mode_ <= 2);
```

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: I exposed this as a param. I'm not too sure how properly expose this param through the GUI, but I can look into it more if required.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: I removed this value. It seems to only be used to surface warnings.

corelib/src/odometry/OdometryCuVSLAM.cpp
Dekempsy4: 1 is allowed, I changed this block to use an assertion.

martincerven:
Do you need special CMAKE flags to enable this? And have you tested it with docker?

matlabbe:
You can look the original PR here https://github.com/introlab/rtabmap/pull/1583#issue-3427606360

Normally if you install `ros-humble-isaac-ros-visual-slam`, when rebuilding rtabmap repo it should detect it automatically if you build with `-DWITH_CUVSLAM=ON`. Beware of some shared library issues described in the same PR.

PR #1626, opened 2025-12-06
support gtsam 4.3 header migration

v4hn:
They recently split this class out in another header [in their 4.3 development branch](https://github.com/borglab/gtsam/pull/2301).

### DIFF
--- a/corelib/src/optimizer/gtsam/GravityFactor.h
+++ b/corelib/src/optimizer/gtsam/GravityFactor.h
@@ -25,6 +25,9 @@
 #pragma once
 
 #include <gtsam/nonlinear/NonlinearFactor.h>
+#if GTSAM_VERSION_NUMERIC >= 40300
+#include <gtsam/nonlinear/NoiseModelFactorN.h>
+#endif
 #include <gtsam/geometry/Pose3.h>
 #include <gtsam/geometry/Unit3.h>
 


matlabbe:
LGTM! thank you

matlabbe:
Looks like kilted and rolling are using a 4.3 version older than this change added 2 weeks ago: https://github.com/borglab/gtsam/commits/develop/gtsam/nonlinear/NoiseModelFactorN.h

They just released 3 weeks ago a new version for ros2: https://github.com/borglab/gtsam/issues/2278

v4hn:
Wonderful. I guess no choice but to wait for ros2 to pull in new versions once things settled down there, because the different development states share the version number.
I packaged it for ROS-O in [one of the builders](https://github.com/v4hn/ros-o-builder/commit/3496806701b25de24c945b1c8dd26fc5abda051d) last week and found this breaking change.
I guess there is no harm in having it lie around depending on your preferences for the master branch.

matlabbe:
Yeah, we are in a weird position:
* Updating it here means we cannot build against gtsam binaries
* Not updating here means we cannot build with latest gtsam build from source

We had a similar issue with g2o library, which doesn't use versions and API gets broken from time to time. It seems that file [gtsam/nonlinear/NoiseModelFactorN.h](https://github.com/borglab/gtsam/pull/2301/changes#diff-b4b6f7052a06c4041ad82ac97af67c8251a6095c77a987490cc7abfbe3281af7) is a new file. I'll check to add a special case detecting that header with cmake and add a preprocessor define to add the include in [GravityFactor.h](https://github.com/introlab/rtabmap/pull/1626/changes#diff-f66b2d9ab9988a5745960a6dae1ac0efcd3533dacabc9b5dfc16afb622fa1c46) file.

matlabbe:
Added support in https://github.com/introlab/rtabmap/commit/6f715b4a8dc3ed1c482160a4feb2a7a9b467c613

PR #1544, opened 2025-06-30
Remove typo in ceres optimizer branch

HO4X:
Fixed the typo in corelib/src/optimizer/OptimizerCeres.cpp#L200

PR #1645, opened 2026-01-25
fix another typo for ceres optimizer

borongyuan:
After enabling Ceres, the compilation failed, and I found that #1544 only fixed half of the typo:>


matlabbe:
Great! thank you

PR #1667, opened 2026-03-06
Add params to filter out poor stereo flow matches

borongyuan:
When using OpenCV's classic Lucas-Kanade optical flow, there are two error measures that can be used to filter out some bad results. A common configuration is to check the minimum eigenvalue of the spatial gradient matrix, which can filter out some features that are not rich enough in texture in advance. This was also a configuration that was previously hardcoded in RTAB-Map. Even if a feature has rich texture, optical flow may still result in an incorrect match. After experiencing some pitfalls, I now prefer to use L1 distance between patches for filtering.
For OpenCV's calcOpticalFlowPyrLK(), ErrorThreshold can generally be set to around 20. For the modified calcOpticalFlowPyrLKStereo(), since the stereo rectification may not be perfect, the error obtained after limiting the optical flow matching direction will generally be larger. Therefore, ErrorThreshold can be set to around 50.
We should probably also expose these params for the optical flow used in VO and visual registration.


matlabbe:
Thanks for this new feature. I was stuck on that PR https://github.com/introlab/rtabmap/pull/1656 the past two weeks, I'll try to take a look at this one this week!

matlabbe:
Looks pretty good, thank you for taking the time to open a PR!

As suggested, I updated the PR with the same parameters for F2F/optical flow, using error threshold of 20 by default. You can find them under `Vis/CorFlow****` group.

I renamed `Stereo/GetMinEigenValues` to `Stereo/UseMinEigenValues`.

PR #1694, opened 2026-04-29
Preserve the attributes of intermediate nodes when retrieving data from the database

borongyuan:
When we reprocess a database containing intermediate nodes, the intermediate nodes are treated as normal nodes. This presents some problems because intermediate nodes typically do not contain data, or only contain sensor data without a signature. After checking, I found that it seems only this one line of fix is ​​needed.


matlabbe:
That should be the right fix, indeed. I didn't try replaying a database with intermediate nodes, glad it works with this little change. Thanks for the PR!

Issue #622, opened 2020-10-27
Intermediate Nodes [question]


NicoNorena:
What is the purpose of the Intermediate nodes?
 In what scenario would it be needed?

matlabbe:
This option has been added to evaluate the SLAM approach on KITTI dataset. Their evaluation tool required to have exactly the same number of poses than the images processed. By default, RTAB-Map won't save all odometry frames, as it should run at lower rate. Enabling that option makes RTAB-Map adding those intermediate poses in the graph, while not saving their data. When exporting the final graph, we have all odometry poses.

NicoNorena:
okay, got it thanks.

PR #1697, opened 2026-05-06
Fixed some issues in processing intermediate node data

borongyuan:
Hi @matlabbe,

Recently, some users have been trying to use RTAB-Map for embodied AI data acquisition. Therefore, we need to create intermediate nodes to record the complete VIO trajectory. Then we can perform offline optimization and export. Based on [the purpose of creating the intermediate node](https://github.com/introlab/rtabmap/issues/622), this usage should be reasonable. However, this configuration doesn't seem to have been used much before, and we encountered some issues during testing.

The most obvious problem is that some redundant nodes are removed when the camera is stationary or when a rehearsal merge is triggered. However, we expect these redundant nodes to be retained when CreateIntermediateNodes is enabled to ensure the integrity of the trajectory. Therefore, these nodes should not be removed from the pose graph, but should be converted into intermediate nodes. The first commit is to fix this behavior.

Then, we want to keep the data of the intermediate node only when IntermediateNodeDataKept is true, otherwise only record its pose. Therefore, I further added the data cleaning section. However, there seems to be a minor issue: RtabmapThread only retains image data when creating intermediate nodes, and discards feature data. Therefore, if we want to use intermediate nodes as normal nodes during reprocessing, we will have to re-detect features. In [the previous commit](https://github.com/introlab/rtabmap/commit/0cf2e45c811aab76117e4d51544c5f5e376c9be5), we actually prohibited the use of intermediate nodes as normal nodes during reprocessing, because by default, intermediate nodes do not have valid data. If there is a need for this in the future, I should further improve this part of the behavior.

### DIFF
--- a/corelib/include/rtabmap/core/Memory.h
+++ b/corelib/include/rtabmap/core/Memory.h
@@ -140,6 +140,7 @@ public:
 			float radius,
 			const std::map<int, Transform> & optimizedPoses,
 			int maxGraphDepth) const;
+	void convertToIntermediate(int locationId);
 	void deleteLocation(int locationId, std::list<int> * deletedWords = 0);
 	void saveLocationData(int locationId);
 	void removeLink(int idA, int idB);
@@ -338,6 +339,7 @@ private:
 	float _rehearsalMaxDistance;
 	float _rehearsalMaxAngle;
 	bool _rehearsalWeightIgnoredWhileMoving;
+	bool _createIntermediateNodes;
 	bool _useOdometryFeatures;
 	bool _useOdometryGravity;
 	bool _rotateImagesUpsideUp;
--- a/corelib/include/rtabmap/core/Rtabmap.h
+++ b/corelib/include/rtabmap/core/Rtabmap.h
@@ -125,6 +125,7 @@ public:
 	void close(bool databaseSaved = true, const std::string & ouputDatabasePath = "");
 
 	const std::string & getWorkingDir() const {return _wDir;}
+	bool isCreateIntermediateNodes() const { return _createIntermediateNodes; }
 	bool isRGBDMode() const { return _rgbdSlamMode; }
 	int getLoopClosureId() const {return _loopClosureHypothesis.first;}
 	float getLoopClosureValue() const {return _loopClosureHypothesis.second;}
@@ -282,6 +283,7 @@ private:
 
 private:
 	// Modifiable parameters
+	bool _createIntermediateNodes;
 	bool _publishStats;
 	bool _publishLastSignatureData;
 	bool _publishPdf;
--- a/corelib/include/rtabmap/core/RtabmapThread.h
+++ b/corelib/include/rtabmap/core/RtabmapThread.h
@@ -66,11 +66,9 @@ public:
 	void clearBufferedData();
 	void setDetectorRate(float rate);
 	void setDataBufferSize(unsigned int bufferSize);
-	void createIntermediateNodes(bool enabled);
 
 	float getDetectorRate() const {return _rate;}
 	unsigned int getDataBufferSize() const {return _dataBufferMaxSize;}
-	bool getCreateIntermediateNodes() const {return _createIntermediateNodes;}
 
 	/**
 	 * Close rtabmap. This will delete rtabmap object if set.
@@ -105,7 +103,6 @@ private:
 	USemaphore _dataAdded;
 	unsigned int _dataBufferMaxSize;
 	float _rate;
-	bool _createIntermediateNodes;
 	double _previousStamp;
 
 	Rtabmap * _rtabmap;
--- a/corelib/src/Memory.cpp
+++ b/corelib/src/Memory.cpp
@@ -117,6 +117,7 @@ Memory::Memory(const ParametersMap & parameters) :
 	_createOccupancyGrid(Parameters::defaultRGBDCreateOccupancyGrid()),
 	_visMaxFeatures(Parameters::defaultVisMaxFeatures()),
 	_visSSC(Parameters::defaultVisSSC()),
+	_createIntermediateNodes(Parameters::defaultRtabmapCreateIntermediateNodes()),
 	_imagesAlreadyRectified(Parameters::defaultRtabmapImagesAlreadyRectified()),
 	_rectifyOnlyFeatures(Parameters::defaultRtabmapRectifyOnlyFeatures()),
 	_covOffDiagonalIgnored(Parameters::defaultMemCovOffDiagIgnored()),
@@ -757,6 +758,7 @@ void Memory::parseParameters(const ParametersMap & parameters)
 	Parameters::parse(params, Parameters::kRGBDCreateOccupancyGrid(), _createOccupancyGrid);
 	Parameters::parse(params, Parameters::kVisMaxFeatures(), _visMaxFeatures);
 	Parameters::parse(params, Parameters::kVisSSC(), _visSSC);
+	Parameters::parse(params, Parameters::kRtabmapCreateIntermediateNodes(), _createIntermediateNodes);
 	Parameters::parse(params, Parameters::kRtabmapImagesAlreadyRectified(), _imagesAlreadyRectified);
 	Parameters::parse(params, Parameters::kRtabmapRectifyOnlyFeatures(), _rectifyOnlyFeatures);
 	Parameters::parse(params, Parameters::kMemCovOffDiagIgnored(), _covOffDiagonalIgnored);
@@ -2413,7 +2415,7 @@ int Memory::cleanup()
 	int signatureRemoved = 0;
 
 	// bad signature
-	if(_lastSignature && ((_lastSignature->isBadSignature() && _badSignaturesIgnored) || !_incrementalMemory))
+	if(_lastSignature && ((_lastSignature->isBadSignature() && _badSignaturesIgnored) || !(_incrementalMemory || _createIntermediateNodes)))
 	{
 		if(_lastSignature->isBadSignature())
 		{
@@ -3040,6 +3042,16 @@ bool Memory::setUserData(int id, const cv::Mat & data)
 		UERROR("Node %d not found in RAM, failed to set user data (size=%d)!", id, data.total());
 	}
 	return false;
+}
+
+void Memory::convertToIntermediate(int locationId)
+{
+	UDEBUG("Converting location %d to intermediate node", locationId);
+	Signature * location = _getSignature(locationId);
+	if(location)
+	{
+		location->setWeight(-1);
+	}
 }
 
 void Memory::deleteLocation(int locationId, std::list<int> * deletedWords)
@@ -4226,8 +4238,8 @@ bool Memory::rehearsalMerge(int oldId, int newId)
 						_rehearsalMaxDistance, _rehearsalMaxAngle);
 				return false;
 			}
-			fullMerge = !isMoving && newS->hasLink(oldS->id());
-			intermediateMerge = !isMoving && !newS->hasLink(oldS->id());
+			fullMerge = !isMoving && (newS->hasLink(oldS->id()) && !_createIntermediateNodes);
+			intermediateMerge = !isMoving && (!newS->hasLink(oldS->id()) || _createIntermediateNodes);
 		}
 		else
 		{
--- a/corelib/src/Memory.cpp
+++ b/corelib/src/Memory.cpp
@@ -3051,6 +3051,14 @@ void Memory::convertToIntermediate(int locationId)
 	if(location)
 	{
 		location->setWeight(-1);
+		if(!_saveIntermediateNodeData)
+		{
+			this->disableWordsRef(locationId);
+			location->removeAllWords();
+			location->sensorData().setFeatures(std::vector<cv::KeyPoint>(), std::vector<cv::Point3f>(), cv::Mat());
+			location->sensorData().clearCompressedData(true, false, false);
+			location->sensorData().clearRawData(true, false, false);
+		}
 	}
 }
 
@@ -4329,7 +4337,14 @@ bool Memory::rehearsalMerge(int oldId, int newId)
 				// just update weight
 				int w = oldS->getWeight()>=0?oldS->getWeight():0;
 				newS->setWeight(w + newS->getWeight() + 1);
-				oldS->setWeight(intermediateMerge?-1:0); // convert to intermediate node
+				if(intermediateMerge)
+				{
+					this->convertToIntermediate(oldS->id());
+				}
+				else
+				{
+					oldS->setWeight(0);
+				}
 
 				if(_lastGlobalLoopClosureId == oldS->id())
 				{
@@ -4340,7 +4355,14 @@ bool Memory::rehearsalMerge(int oldId, int newId)
 			{
 				int w = newS->getWeight()>=0?newS->getWeight():0;
 				oldS->setWeight(w + oldS->getWeight() + 1);
-				newS->setWeight(intermediateMerge?-1:0); // convert to intermediate node
+				if(intermediateMerge)
+				{
+					this->convertToIntermediate(newS->id());
+				}
+				else
+				{
+					newS->setWeight(0);
+				}
 			}
 		}
 	}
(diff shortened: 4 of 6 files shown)


matlabbe:
I wonder what is the difference with:
```
rtabmap \
   --RGBD/LinearUpdate 0 \
   --RGBD/AngularUpdate 0 \
   --Rtabmap/CreateIntermediateNodes true \
   --Mem/IntermediateNodeDataKept true
```
which would I think provide already what you need. We could however add an option in `DBReader` to treat intermediate nodes as normal nodes if we want to reprocess offline all of them as normal nodes (and re-detecting features in them).

But yeah, the only practical usage so far of intermediate nodes was for KITTI evaluation dataset:
https://github.com/introlab/rtabmap/blob/f04067fa16c7699a8b934644a7eda8bc9b188578/docker/jfr2018/run_kitti_datasets.sh#L114
which was used in combination with `--RGBD/LinearUpdate 0 --RGBD/AngularUpdate 0`.

However, the two lines with `intermediateMerge` here can be confusing:
https://github.com/introlab/rtabmap/blob/f04067fa16c7699a8b934644a7eda8bc9b188578/corelib/src/Memory.cpp#L4314-L4332

While the second "convert to intermediate node" could signal Rtabmap to not add the new node to the local graph, the first one won't remove the previous node from the graph. Only the second path (using default value from `Mem/RehearsalIdUpdatedToNewOne=false`) has been really supported. Another assumption was that the "new" intermediate node been also removed because of "too small motion" on Rtabmap side, though it seems to be broken by:
https://github.com/introlab/rtabmap/blob/f04067fa16c7699a8b934644a7eda8bc9b188578/corelib/src/Rtabmap.cpp#L1535-L1539
that was added 3 weeks ago: https://github.com/introlab/rtabmap/commit/9cbe84e44533f9e14d66118e6c69ba67c291cf38

My 2 cents tell me that we should not convert to intermediate node in `Memory::RehearsalMerge()`, because that function should only be called with non-intermediate nodes (which have also visual words in the vocabulary). Another issue is that the behavior in DBReader that we [added](https://github.com/introlab/rtabmap/commit/0cf2e45c811aab76117e4d51544c5f5e376c9be5) recently about replaying intermediate nodes won't work in that situation. I'll need to think about it.

borongyuan:
> rtabmap \
>    --RGBD/LinearUpdate 0 \
>    --RGBD/AngularUpdate 0 \
>    --Rtabmap/CreateIntermediateNodes true \
>    --Mem/IntermediateNodeDataKept true

The parameters we used originally were:
```
Kp\BadSignRatio = 0.0
Mem\RehearsalSimilarity = 1.0
Mem\STMSize = 1
RGBD\AngularUpdate = 0
RGBD\LinearUpdate = 0
RGBD\MaxLocalRetrieved = 0
RGBD\ProximityBySpace = false
Rtabmap\CreateIntermediateNodes = true
Rtabmap\DetectionRate = 1.0
Rtabmap\MaxRetrieved = 0
Rtabmap\MemoryThr = 2
```
This is basically the same as the parameters for data recorder mode. The main difference is that we still keep the detection rate at 1 Hz, but create intermediate nodes. Then we reprocess the recorded data, including loop closure detection and merging multiple databases. This way, there will be no invalid nodes in the recorded data, but there will still be some redundant nodes. Even if the camera is stationary, a new normal node will be added every second. This is because the STM function has also been disabled. If AngularUpdate and LinearUpdate are set and rehearsal is enabled, these redundant nodes can be filtered out, but they will become invalid nodes with a weight of -9 and be removed from the pose graph. What we really want is to record VIO data while running online, retain the functionality of STM, convert redundant nodes into intermediate nodes, and disable online loop closure detection. Therefore, these changes are necessary to achieve the desired effect. Our goal is to ensure that the system behaves consistently across various configuration combinations, regardless of whether CreateIntermediateNodes is enabled. When CreateIntermediateNodes is enabled, no nodes should be removed from the pose graph or invalidated; instead, they should all be converted into intermediate nodes.

After these changes, there's still a minor issue. We can't completely disable online loopback detection yet. In the data recorder mode configuration, loop closure detection is indirectly disabled by disabling feature detection. However, our feature data is provided by VIO, and we need to record this data. We will not re-perform feature detection during reprocessing. Even if we reduce MemoryThr to its minimum, the online loop closure detection will still work. We temporarily avoid generating loop closures online by modifying LoopThr, but it only rejects loop closures, rather than disabling this feature.

> My 2 cents tell me that we should not convert to intermediate node in `Memory::RehearsalMerge()`, because that function should only be called with non-intermediate nodes (which have also visual words in the vocabulary).

If we want to ensure that all nodes are preserved when CreateIntermediateNodes is enabled and rehearsal can still be used, then nodes cannot be deleted during merge, but must be converted into intermediate nodes. But I'm also a little uncomfortable with this part of the change. We now need both the Rtabmap and Memory classes to obtain the CreateIntermediateNodes parameter. However, from a code structure perspective, each parameter should ideally belong to only one level.

matlabbe:
What is the purpose of keeping loop closure detection in that scenario? Because with `Rtabmap\MemoryThr=2`, it is basically disabling loop closure detection.
> This is basically the same as the parameters for data recorder mode. The main difference is that we still keep the detection rate at 1 Hz, but create intermediate nodes. 

If it is only to keep features that were extracted in VIO, I think we could make a small modification with `Kp/MaxFeatures=-1` that will still forward input features in the new Signature. No intermediate nodes needed, just do what Data recorder mode does, but save the input visual features (keypoints/descriptors). That would maybe address:
>What we really want is to record VIO data while running online, retain the functionality of STM, convert redundant nodes into intermediate nodes, and disable online loop closure detection. 

I am not sure to understand which part of "retain the functionality of STM" you want to keep during recording. Also what does "redundant nodes" mean to you (e.g., when not moving and/or rehearsal is triggered)?

I created another branch with the changes I think could work for your workflow. Am I missing something?
https://github.com/introlab/rtabmap/pull/1698

borongyuan:
Okay, here's what the current needs are in the data acquisition field. Some users now want to build data acquisition devices similar to [FastUMI](https://fastumi.com), or other types of devices. They need to record the trajectory of these devices. Currently, various positioning solutions are being explored, including those using consumer-grade devices like the HTC Vive, more expensive motion capture systems, and SLAM. SLAM is more flexible than other solutions and does not require a base station or a specific reference. However, its accuracy is an issue. VIO will accumulate errors, which can be eliminated through loop closure detection. After communicating with some friends in this field, most of them believe that recording data online first and then optimizing it offline is the most suitable approach. If we use the regular SLAM running mode and record the trajectory online, there will be some problems. If the VIO trajectory is recorded, it will be smooth and continuous, but without graph optimization, there will be accumulated errors. If the trajectory under the map frame is recorded, the pose will jump during each loop closure detection and optimization. Moreover, online optimization can only use historical information and cannot use future data. Therefore, when running online, we only need to record VIO data and do not perform loop closure detection. Using reprocess for offline processing has other advantages. Some applications may need to record the trajectories of two or more devices simultaneously and then merge them into the same coordinate system. As long as there are shared viewing areas among these trajectories, RTAB-Map can automatically merge the databases. If each node recorded complete data, the amount of data would be enormous, which is unnecessary. Therefore, we need to use intermediate nodes to preserve their pose but not their data. We want to retain the rehearsal feature during online operation, so that new data will also be converted into intermediate node if it is similar to the most recent data in STM.

matlabbe:
> If the VIO trajectory is recorded, it will be smooth and continuous, but without graph optimization, there will be accumulated errors.

Yes

> If the trajectory under the map frame is recorded, the pose will jump during each loop closure detection and optimization.

Yes

> Moreover, online optimization can only use historical information and cannot use future data.

Yes, but offline optimization is kinda the same thing too. In theory, if all loop closures that can be detected online are detected, the result would be similar than doing offline optimization. In practice, the problem is that online SLAM may miss some loop closures that only after detecting future loop closures it could have detected past ones... like backtracking on the past trajectory to check all loop closures that were missed. It is even more true with RTABMap's memory management enabled, where large parts of the map would be in LTM.

>  Therefore, when running online, we only need to record VIO data and do not perform loop closure detection. 

Setting up RTAB-Map in "data recording" mode can do that.

> As long as there are shared viewing areas among these trajectories, RTAB-Map can automatically merge the databases.

I think it is an important point why it is still preferable to have online loop closure detection enabled, otherwise, we don't have any feedback if the data recorded can actually close the loop closures offline. I agree that if we want to keep all VIO data at the same time, how the intermediate nodes are handled in the current version are not working. In https://github.com/introlab/rtabmap/pull/1698, you can use `Rtabmap/CreateIntermediateNodes` with `Mem/IntermediateNodeDataKept` to record all VIO data while doing online loop closure detection at 1 Hz. That PR also adds an option `-pub_inter_as_normal` to `rtabmap-reprocess` to reprocess all VIO nodes as normal nodes, so even possible to do loop closure detection at 30 Hz (if VIO rate it is) offline. I also made a fix when `RGBD/LinearUpdate` and `RGBD/AngularUpdate` triggers with Rehearsal that the nodes are not disconnected from the graph like before, but converted to intermediate nodes instead.

> If each node recorded complete data, the amount of data would be enormous, which is unnecessary.

If we want to record all VIO data **without** online loop closure detection only when the camera is not moving, I would recommend to setup RTAB-Map in "data recording" mode and set `RGBD/AngularUpdate` and/or `RGBD/LinearUpdate` to a value that will discard data (with `Mem/NotLinkedNodesKept=false`) if not moving enough.

If we want to record all VIO data **with** online loop closure detection only when the camera is not moving, well, the only closest config is like described above with `Rtabmap/CreateIntermediateNodes` and `Mem/IntermediateNodeDataKept` (with this PR https://github.com/introlab/rtabmap/pull/1698), but nodes (intermediate or not) when not moving won't be discarded.

> Therefore, we need to use intermediate nodes to preserve their pose but not their data. We want to retain the rehearsal feature during online operation, so that new data will also be converted into intermediate node if it is similar to the most recent data in STM.

It seems this mode is `Rtabmap/CreateIntermediateNodes=true`, `Mem/IntermediateNodeDataKept=false`, converting rehearsed nodes to intermediate nodes under  `RGBD/AngularUpdate` and `RGBD/LinearUpdate`. That is pretty close to what https://github.com/introlab/rtabmap/pull/1698 does now, at one exception is that the converted node to intermediate could still have the raw data (unlike other intermediate nodes). That what you did in that function:
```cpp
void Memory::convertToIntermediate(int locationId)
{
	UDEBUG("Converting location %d to intermediate node", locationId);
	Signature * location = _getSignature(locationId);
	if(location)
	{
		location->setWeight(-1);
		if(!_saveIntermediateNodeData)
		{
			this->disableWordsRef(locationId);
			location->removeAllWords();
			location->sensorData().setFeatures(std::vector<cv::KeyPoint>(), std::vector<cv::Point3f>(), cv::Mat());
			location->sensorData().clearCompressedData(true, false, false);
			location->sensorData().clearRawData(true, false, false);
		}
	}
}
```
Yeah, we could add this.

I think I've misread this part in the PR description:
> RtabmapThread only retains image data when creating intermediate nodes, and discards feature data. Therefore, if we want to use intermediate nodes as normal nodes during reprocessing, we will have to re-detect features.

I've read that the features provided by VIO were not kept in intermediate nodes when `Mem/IntermediateNodeDataKept=true` (which is fixed in https://github.com/introlab/rtabmap/pull/1698). From what you are saying, we should remove that
https://github.com/introlab/rtabmap/blob/afe8aadff1e5b80e2032da9313b5f6ff208d3ec7/corelib/src/RtabmapThread.cpp#L559
I don't have a particular reason why we should keep it.


So back to this PR, I don't think Rtabmap and Memory need to know about `Rtabmap/CreateIntermediateNodes` parameter. For 
```cpp
fullMerge = !isMoving && (newS->hasLink(oldS->id()) && !_createIntermediateNodes);
intermediateMerge = !isMoving && (!newS->hasLink(oldS->id()) || _createIntermediateNodes);
```
not having a direct link to oldS is already a proxy to know there are intermediate nodes. In Rtabmap, the forget step under `(smallDisplacement || tooFastMovement)` would be already handled by Rehearsal for the `smallDisplacement`. For `tooFastMovement`, we could convert to intermediate node right away when it is computed, so that the rest of the code assumes the node is intermediate (which changes slightly how the graph's constraints and poses are updated). So for
https://github.com/introlab/rtabmap/blob/afe8aadff1e5b80e2032da9313b5f6ff208d3ec7/corelib/src/Rtabmap.cpp#L1569-L1573
we may do this instead:
```cpp
// like for too small motion, intermediate nodes are already "discarded" (so ignore this check for intermediate nodes)
tooFastMovement = signature->getWeight() >= 0 &&
 		((_rgbdLinearSpeedUpdate>0.0f && uMax3(fabs(odomVelocity[0]), fabs(odomVelocity[1]), fabs(odomVelocity[2])) > _rgbdLinearSpeedUpdate) || 
 		(_rgbdAngularSpeedUpdate>0.0f && uMax3(fabs(odomVelocity[3]), fabs(odomVelocity[4]), fabs(odomVelocity[5])) > _rgbdAngularSpeedUpdate));
if(tooFastMovement && signature->hasLink(<<<with  an intermediate node>>>)) {
    _memory->convertToIntermediate(signature->id());
    tooFastMovement = false; // intermediate nodes are removed from the local graph automatically, but we keep them linked in the database
}
```

So what I will do is to merge your branch in [#1698](https://github.com/introlab/rtabmap/pull/1698) for `void Memory::convertToIntermediate(int locationId)` and make the changes above.

borongyuan:
> Yes, but offline optimization is kinda the same thing too. In theory, if all loop closures that can be detected online are detected, the result would be similar than doing offline optimization. In practice, the problem is that online SLAM may miss some loop closures that only after detecting future loop closures it could have detected past ones... like backtracking on the past trajectory to check all loop closures that were missed. It is even more true with RTABMap's memory management enabled, where large parts of the map would be in LTM.

Yes, online and offline loop closure detection should yield similar or even identical results. Our concern is that only when exporting data from a database are the complete pose graph and all added loop closure constraints utilized. When exporting using rtabmap-export, a global optimization is performed by default, resulting in a smooth trajectory, which is theoretically the most accurate result (possibly second only to global BA). We are not concerned with whether these loop constraints were added online or offline. Another reason we want to disable online loop closure detection is to save computing resources. For example, if we connect four cameras to a Jetson and run four SLAM instances simultaneously, our VIO is already lightweight enough, so running four instances at the same time should be fine. However, if there are also four RTAB-Map backends running, then it's hard to say.

> Setting up RTAB-Map in "data recording" mode can do that.

Using rtabmap-dataRecorder is actually the most resource-efficient because it doesn't run the main RTAB-Map process. However, we are running through ROS, and there is no corresponding ROS node for this part. Therefore, we can only use the data recording mode, which still runs the RTAB-Map main process, but we can try to turn off the functions we don't need. Our custom VIO will package the data in rtabmap_ros/RGBDImage and pass it to rtabmap_slam, which includes compressed RGB+depth images, features, and global descriptor. If all the data were recorded, the database would become too large. Storage usage itself isn't the main issue. The main problem is that it would drastically slow down reprocessing. The VIO outputs data at a frequency of 20 Hz, and recording just 10 minutes of data will generate tens of thousands of nodes, which is completely different from the previous running mode. If multiple datasets are merged and improperly configured, reprocessing time can reach several hours. Therefore, we want to convert most nodes into intermediate nodes while recording data and fix related issues.

> I've read that the features provided by VIO were not kept in intermediate nodes when `Mem/IntermediateNodeDataKept=true` (which is fixed in #1698). From what you are saying, we should remove that

I've also noticed this part. At this stage, we skip intermediate nodes during reprocessing, so it doesn't matter whether they contain data. Therefore, I haven't done a full test on the function of Mem/IntermediateNodeDataKept. Ideally, it should work exactly as expected (i.e., either retaining or deleting the data completely). This facilitates subsequent research, allowing for the recording of complete data that, when used as intermediate nodes, does not affect performance, but can also be restored to normal node use when necessary.

> So what I will do is to merge your branch in [#1698](https://github.com/introlab/rtabmap/pull/1698) for `void Memory::convertToIntermediate(int locationId)` and make the changes above.

Okay, I don't see any problem with this modification. We can test and verify it later.

matlabbe:
Merged to https://github.com/introlab/rtabmap/pull/1698, I'll doublecheck tonight if all the workflows work as expected and merge it.

matlabbe:
https://github.com/introlab/rtabmap/pull/1698 has been merged.

PR #1728, opened 2026-07-02
Faster graph zoom by using a snapshot-preview

elidiot18:
Hi,
Browsing the graph of a large database can be tedious because each tiny bit of zoom requires to rerender all visible objects. In this commit I display a raster snapshot temporarily and render the real canvas only when the user has finished zooming (400ms need to pass without a new mouse wheel).
Note that I also changed the zooming anchor to mouse-centered because it feels more natural to me but this is a personal choice. I also updated the shift+click zooming.
It makes it easier to work with large dbs but introduces some pixelisation for normal size dbs, maybe it should only be activated if the db is detected as "large".
Note that the same should could also be done for rotating the graph.

https://github.com/user-attachments/assets/9ff6ccd2-4b77-4687-9125-b1929388e8bf


matlabbe:
That looks great! I added a right-click menu option to enable the rastering. On my computer, the rendering is smooth for a 12000 nodes map I could test with. Maybe on an 100K nodes I would start to see lags as you experienced. By default the rastering is disabled, but can be enabled with this option (will persist between UI sessions if you save the UI settings).

image
Screenshot from 2026-07-05 12-00-02

The shift-click zoom feels also better to use.

Issue #1671, opened 2026-03-15
Improve detect more loop closures tool to avoid overpowering odometry contribution in local graph optimization


matlabbe:
The issue is that when using `rtabmap-detectMoreLoopClosures` tool (or doing it in the UI) with many iterations, it could add many loop closures between nodes that are very close in the graph (e..g, nodes taken a couple of seconds apart). Normally, for a couple of nodes in the graph, the odometry constraints should be more accurate than loop closure constraints. Thus keeping locally consistent optimization is already implicitly handled by smaller covariance for odometry links and larger covariance for loop closure links. However, even if loop closure links have generally higher covariance, adding many of them on consecutive nodes could bend the odometry constraints (maybe because the odometry was not confident enough). The idea would be to add a parameter that will make detectMoreLoopClosures ignore nodes close in the graph (in terms of number of nodes between them or in time) for potential loop closures.

In summary, currently doing many iterations in that tool can make the overall map worst because loop closure constraints overpower the odometry constraints locally.

Original discussion: http://official-rtab-map-forum.206.s1.nabble.com/More-loops-isn-t-always-better-td11560.html

PR #1729, opened 2026-07-07
Multi component view

elidiot18:
Hi,

I tried to implement a multi component framework for the graph canvas. Useful when a reprocess was not able to join all pieces, or when user wants to get rid of a short sequence causing drift while keeping track of the removed piece; also I have other ideas for follow up impovements based on that feature. A small feature allowing to jump on the node corresponding to the displayed image was also implemented.

edit: draft, needs more work.

https://github.com/user-attachments/assets/786b6a41-88be-48be-af2a-f7992022d844

PS: Note that for integrating to the main project with minimal changes and keeping the "optimization from fromId" philosophy, I kept only one QGraphicsView, which means that going from one component to the other implies to first backup all current parameters, then change component and restore all parameters of the new component. This way, the new feature is mainly some new code plugged on top rather than a redesign. Maybe it would be cleaner though to dedicate one QGraphicsView per component instead ?

### DIFF
--- a/.github/actions/install-macos-source-deps/action.yml
+++ b/.github/actions/install-macos-source-deps/action.yml
@@ -124,7 +124,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ runner.temp }}/deps-stage/depthai
-        key: depthai-2.32.0-usrlocal-${{ inputs.os }}-${{ hashFiles('.github/actions/install-macos-source-deps/patches/depthai-2.32.0-hunter-macos.patch') }}-${{ steps.depver.outputs.hash }}
+        key: depthai-2.32.0-usrlocal-nocv-${{ inputs.os }}-${{ hashFiles('.github/actions/install-macos-source-deps/patches/depthai-2.32.0-hunter-macos.patch') }}-${{ steps.depver.outputs.hash }}
 
     - name: Build depthai
       if: steps.cache-depthai.outputs.cache-hit != 'true'
@@ -161,6 +161,7 @@ runs:
           -DDEPTHAI_ENABLE_CURL=OFF \
           -DDEPTHAI_BUILD_TESTS=OFF \
           -DDEPTHAI_BUILD_EXAMPLES=OFF \
+          -DDEPTHAI_OPENCV_SUPPORT=OFF \
           -DCMAKE_PREFIX_PATH="$(brew --prefix zlib)"
         "$CMAKE3" --build build -j$NPROC
         DESTDIR="$STAGE" "$CMAKE3" --install build
--- a/.github/workflows/cmake-macos.yml
+++ b/.github/workflows/cmake-macos.yml
@@ -22,23 +22,33 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        build_name: [macos-sequoia-intel, macos-sequoia-apple-silicon, macos-tahoe-intel, macos-tahoe-apple-silicon]
+        build_name: [macos-sequoia-intel, macos-sequoia-apple-silicon, macos-tahoe-intel, macos-tahoe-apple-silicon, macos-tahoe-intel-cv5, macos-tahoe-apple-silicon-cv5]
         include:
         - build_name: macos-sequoia-intel
           os: macos-15-intel
+          cv: opencv@4
         - build_name: macos-sequoia-apple-silicon
           os: macos-15
+          cv: opencv@4
         - build_name: macos-tahoe-intel
           os: macos-26-intel
+          cv: opencv@4
         - build_name: macos-tahoe-apple-silicon
           os: macos-26
+          cv: opencv@4
+        - build_name: macos-tahoe-intel-cv5
+          os: macos-26-intel
+          cv: opencv
+        - build_name: macos-tahoe-apple-silicon-cv5
+          os: macos-26
+          cv: opencv
     steps:
     - uses: actions/checkout@v4
 
     - name: Install Brew Dependencies
       run: |
         # Update brew and install from Brewfile if present, or specific packages
-        brew install pcl opencv octomap pdal yaml-cpp librealsense libfreenect libusb zlib libomp suite-sparse ceres-solver
+        brew install pcl octomap pdal yaml-cpp librealsense libfreenect libusb zlib libomp suite-sparse ceres-solver ${{ matrix.cv }}
 
     - name: Install Source Dependencies
       # Build (and per-dependency cache) the source-only deps not available from
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ SET(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake_modules")
 #######################
 SET(RTABMAP_MAJOR_VERSION 0)
 SET(RTABMAP_MINOR_VERSION 23)
-SET(RTABMAP_PATCH_VERSION 8)
+SET(RTABMAP_PATCH_VERSION 9)
 SET(RTABMAP_VERSION
   ${RTABMAP_MAJOR_VERSION}.${RTABMAP_MINOR_VERSION}.${RTABMAP_PATCH_VERSION})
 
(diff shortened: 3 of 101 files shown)


elidiot18:
Hi,

Now should be ready for PR although more natural extensions could be implemented.

**Added features and notes about them**:
- a `GraphComponent` struct which keeps information about the different graph components, some are ui config like flavour optimization and some are about the canvas view
- switching from one component to the other preserving as many things as possible
- one component has to be selected as _main_, and 2D map/optimized mesh can only be generated/edited for this component
- optimized poses are cached in the `GraphComponent `struct but only the main component can have them saved in the db
- an option to zoom around the node associated to an image
- nodes in the graph can now be selected and this can lead to several useful features, only some of which I have implemented
- detectmoreloops can be run only within the selected nodes to optimize a certain region
- components can be merged automatically using user node selection, thanks to a new dedicated tool (_Merge components using selected nodes_)

Natural follow ups would be
- add more "only selected nodes" checkboxes, for instance to export part of the db, or generate clouds only in a certain easy-to-select area, etc.
- add icons for the mouse mode in the canvas instead of radiobuttons
- more serious: implement an intra-db merge tool as I suggested in [this thread](http://official-rtab-map-forum.206.s1.nabble.com/Appending-new-scans-on-top-of-a-large-db-td11604.html), this is one of the main motivations for these new features actually

Note that I would totally understand if you find that those are too radical changes / complications in the DatabaseViewer in which case I'll be happy to keep them in my personal fork !

matlabbe:
Thanks for the PR with these new features. The changes in DatabaseViewer are relatively large, may take some time to review. I like the components approach and that we can zoom to a node in the graph view from the image of it. I didn't see yet your latest changes, but I wonder if having dozen of components would make the "tabs" too cluttered... like tabs on a browser. Keeping only one GraphViewer is simpler though, because many actions (like Export Poses or Export Clouds) rely on the current state of the GraphView.

elidiot18:
Hi Mathieu,

Thanks for your answer. Here is a video that sums up the new changes. The other changes in the code are mainly fixes and polishing for the initial commit (dealing with optimization flavor and rotations, optimized poses).

https://github.com/user-attachments/assets/13c3be31-f45d-4ce2-a423-1471e628d35e

I generated "too many components" on the video so that you can see what it looks like, to me it looks ok and anyway there should never be dozens of components right ? I originally used a slider but found it less natural to use.

Totally fine that it will take some time to review, I expected that :)
I tried to make the changes as minimal as possible so hopefully it won't be too painful.
I will gladly add more comments if it can be of any help to make the review smoother.

About the one GraphView, yes I eventually leaned towards that because it could integrate more naturally to the existing philosophy. Having a "main" component is also going in that direction: only that component can have specific things saved in the db like 2D map of optimized poses.

matlabbe:
Thanks for the demo. There are many features I would want to get in. Would you let me refactor it? I have a couple of questions before doing it:
1. What is the purpose of the "main" component? There are stuff about 2D optimized map or graph, and mesh. I think as soon as components are created, we should disable/clear 2D optimized map/graph and optimized mesh file menu actions. The problem is that all nodes contained in local optimized graph should be in Working Memory. To know however which component is the "main" one, the default map used by the robot, we could add a `*` like you did but cannot be changed manually. If we split the main graph, would need to re-use same conditions that the robot would use on loading to define the new main (based on `RGBD/OptimizeFromGraphEnd` parameter). All File actions would be done on the current selected component instead of main.
2. For the "Zoom to corresponding node", I would remove that action from ImageView (that widget is used in other contexts that don't involve a graph view). I think we could automatically zoom to current node when one of the main slider changes. Would there be a case in your workflow you wouldn't want automatically zoom in graph all the time when browsing images? My guess is if the consecutive images are from different components, it may add time to browse the images (component has to be re-optimized/shown). I could had a checkbox under GraphViewer (or a Gui settings) for "auto-zooming". The auto-zoom could also be extended to the constraints browsing. 
3. I would rename the components tab name from `Component 1`, `Component 2` to `A-435`, `B-231`, `*C-1054`... (so we can see more of them without scrolling and quickly reference them by size). 
4. I would also show the number of nodes selected in the tab's name: `A-435(12)`, `B-231(20)`, `C-1054` (no node selected)
5. first time we click on a component (or when we split the graph, causing a re-optimization), we should do a fit to view.
6. Don't know if we could support "ctrl"/"shift" selection to select/unselect more or less nodes
7. Add option in the "Refine links", or "Add more loop closures" or "Regenerate local grids" or "Update covariances" to do it only on selected nodes. Add menu action like "Clear selection (# selected)".
8. I could check to add a solution for avoiding re-optimizing/re-zooming after a manual graph change.
9. When switching components with a lot of nodes, there is some lag. It seems graph optimization happens again and/or the the 2d occupancy grid map is re-generated and/or re-creating all the graphics view nodes/links. For the 2D occupancy grid and optimized poses, we may be able to cache them (I think that was the commented code in GraphComponent was about).

Minor bugs:
* Components that were split, are not merged again when rejecting all graph changes so far.
* Qt warning on start:
  ```
  QObject::connect: Incompatible sender/receiver arguments
        QSpinBox::editingFinished() --> rtabmap::DatabaseViewer::spinBoxOptimizationsFromValueChanged(int)
  ```
* Fatal error when selecting "Local Optimized" flavor (after splitting the original map):
  ```
  [FATAL] (2026-07-12 11:32:19.036) Link.cpp:166::merge() Condition (to_ == link.from()) not met! [merging this=2851->2869 to link=2870->2871]
  ```

In terms of timing, I have a couple of free days in 2 weeks, I could look at implementing the above and merge it.

elidiot18:
Hi Mathieu,

Thank you for your first review !

> Would you let me refactor it?

Of course

> 1. What is the purpose of the "main" component? 

The goal was to elect one component that's allowed to have its optimized poses and optimized mesh / 2D map saved in db. Otherwise it would overwrite the db at each tab change.

> I think as soon as components are created, we should disable/clear 2D optimized map/graph and optimized mesh file menu actions. 

Sure, I thought that introducing this "main component thing" was less brutal and I wanted my changes to be non destructive. But if you feel that's the right workflow then absolutely.


> The problem is that all nodes contained in local optimized graph should be in Working Memory.[...] All File actions would be done on the current selected component instead of main.

I missed that subtlety,then maybe either the multi component mode should be active, or only one component in the classic mode. I tried to integrate what I wanted to add given my understanding of rtabmap and the dbviewer but obviously there are some things that I'm missing so my pr should be fixed accordingly!

> 2. For the "Zoom to corresponding node", I would remove that action from ImageView (that widget is used in other contexts that don't involve a graph view). I think we could automatically zoom to current node when one of the main slider changes. Would there be a case in your workflow you wouldn't want automatically zoom in graph all the time when browsing?

I think this would work. The issue would be changing component as you noticed, unfortunately when one of the components is very large, switching within the small ones is quite long. I'll try to have a look at why this is the case but probably there is an easy fix to that.


> 3. I would rename the components tab name from Component 1, Component 2 to A-435, B-231, *C-1054... (so we can see more of them without scrolling and quickly reference them by size). 
>
> 4. I would also show the number of nodes selected in the tab's name: A-435(12), B-231(20), C-1054 (no node selected)

Sure everything I proposed is more of a suggestion 

> 
> 5. first time we click on a component (or when we split the graph, causing a re-optimization), we should do a fit to view.

Yes !

> 6. Don't know if we could support "ctrl"/"shift" selection to select/unselect more or less nodes

I tried but failed, lack of experience with Qt I believe :) I'd love if this is possible.

> 7. Add option in the "Refine links", or "Add more loop closures" or "Regenerate local grids" or "Update covariances" to do it only on selected nodes. Add menu action like "Clear selection (# selected)".

Agreed, I wanted your opinion on the selection feature before implementing all this myself :)



> 8. I could check to add a solution for https://github.com/introlab/rtabmap/issues/1671 at the same time

I'm confused, isn't the new parameter `minimumGraphDistance` the solution you implemented ? I thought this issue was solved. Although I made some modifications that I could also propose as PR.

> 9. When switching components with a lot of nodes, there is some lag. It seems graph optimization happens again and/or the the 2d occupancy grid map is re-generated and/or re-creating all the graphics view nodes/links. For the 2D occupancy grid and optimized poses, we may be able to cache them (I think that was the commented code in GraphComponent was about).

Yes I implemented this at first when I was generating maps for all components but then I thought it was easier to just allow it for the main one. Currently updateGraphView is called when changing component because they are re optimized according to the optimization flavour cached parameter. But maybe cached optimized poses can be used by default on a component switch.

> 
> 
> Minor bugs:
> 
> * Components that were split, are not merged again when rejecting all graph changes so far.
> 
> * Qt warning on start:
> 
>  > 
>   QObject::connect: Incompatible sender/receiver arguments
> 
>         QSpinBox::editingFinished() --> rtabmap::DatabaseViewer::spinBoxOptimizationsFromValueChanged(int)
> 
>   > 
> * Fatal error when selecting "Local Optimized" flavor (after splitting the original map):
> 
>  > 
>   [FATAL] (2026-07-12 11:32:19.036) Link.cpp:166::merge() Condition (to_ == link.from()) not met! [merging this=2851->2869 to link=2870->2871]
> 
>   > 
> 
> 
> In terms of timing, I have a couple of free days in 2 weeks, I could look at implementing the above and merge it.

Thanks for finding those bugs. I can also help with fixing the issues you mentioned, but I can't know for sure if I do it the way you would.

Best
Pierre

matlabbe:
8. I miss-linked (I edited my post), though could not find the issue now, what I've meant: I could check to add a solution for avoiding re-optimizing/re-zooming after a manual graph change.

Unless you would want sooner some of the suggestions I've made, you can look at implementing them. I won't really have time to work on this for the next two weeks.

matlabbe:
I refactored the PR following my suggestions. I made all menu actions more aware of the current component shown, and the nodes selected as well (for actions having a dialog for settings, some have new options related to that). For example, "Export database" can extract a component or selected nodes.

For the "Main" component, that is the component representing the last nodes in WM and local graph. If this component is not focused (a sub component is shown), the local graph flavor cannot be selected (disabled) and options changing optimized poses are also disabled (regenerating mesh or 2d map).

I changed the way components are discovered, using `Optimizer::getConnectedGraph()` so that landmark links linking multiple maps can be used to assemble in same component. 

Visually, the components show the number of nodes and the number of selected nodes for convenience. 

For the "Zoom to node" functionality, I moved it next to the sliders. The node ID label is now a button. It now centers the node without zooming.

I think I didn't break the original functionalities. I added new functionalities to add/reject/refine constraints faster without having the graph view either re-zoom automatically or the graph re-optimized all the time. That would make it a lot faster to make manually edits and to avoid losing context on every change.

elidiot18:
This is great :)

PR #1733, opened 2026-07-09
Fixed some issues related to intermediate nodes and retrieval/immunization

borongyuan:
Hi @matlabbe,

We found that when intermediate node was enabled and data recorder mode was used, WM continued to grow even with Rtabmap/MemoryThr set to 1 or 2. This is mainly because the size of WM was used directly when calculating the number of immunized nodes. Since there are intermediate nodes in WM, the number of maxLocalLocationsImmunized is greatly amplified. When counting the number of immunized nodes later, intermediate nodes are skipped, which results in more nodes being immunized and unable to be transferred to LTM. In the previous fix, the method for calculating the size of WM was not updated, so you also mentioned in the PR #1687 that `Rtabmap/MemoryThr` needs to be manually scaled.

I've added two fixes this time: one is the outer layer fix in Rtabmap.cpp, and the other is the inner layer fix in Memory.cpp. When `RGBD/LocalImmunizationRatio`, `RGBD/MaxLocalRetrieved`, and `Rtabmap/MaxRetrieved` are all 0, the retrieval/immunization part can actually be skipped entirely. It is safest for us to skip the outer layer before the inner fix is ​​fully evaluated and validated. Then I added the getWorkingMemSize() method to the Memory class, which allows ignoreIntermediateNodes when calculating the size. The previous code actually contained a comment for getWorkingMemSize(), so adding this method would be the most appropriate solution.
https://github.com/introlab/rtabmap/blob/e321999b1527047dbf52f32d4ad290ff3492ac30/corelib/src/Rtabmap.cpp#L2229
WM contains a virtual node, and after consideration, I think it should not be included when calculating the size. This will make subsequent processing simpler. The only virtual node is inserted at the end of Memory::loadDataFromDb(), which is inside Memory::init(). Therefore, during system operation, WM should always contain a unique virtual node. This implementation should be safe. If you feel there is a risk, we can add some assertions. I have carefully considered whether intermediate nodes should be ignored when obtaining the size of WM in various places, but I may still make some mistakes, so please confirm during the review.

Additionally, there may be an issue with the following call to Memory::reactivateSignatures().
https://github.com/introlab/rtabmap/blob/e321999b1527047dbf52f32d4ad290ff3492ac30/corelib/src/Rtabmap.cpp#L2630-L2640
When the second parameter of Memory::reactivateSignatures() is 0, it does not mean that there is no reactivation, but rather that there is unlimited reactivation. Since reactivatedIds is empty in our configuration, RETRIEVAL 3/3 will be skipped, so this method will not be called. However, this part may need to be checked and fixed later.

I need to submit a PR with existing modifications first, because I see that you are adding support for OpenCV 5, and the modifications also involve this part of the code. Submitting earlier allows for earlier review and merging, reducing the workload of handling conflicts later.

### DIFF
--- a/corelib/src/Rtabmap.cpp
+++ b/corelib/src/Rtabmap.cpp
@@ -2262,7 +2262,7 @@ bool Rtabmap::process(
 	// no need to do retrieval or immunization of locations if memory management
 	// is disabled and all nodes are in WM.
 	// Also skip memory mangement on intermediate nodes
-	if(!(_memory->allNodesInWM() && maxLocalLocationsImmunized == 0) && signature->getWeight()>=0)
+	if(!((_memory->allNodesInWM() || _maxRetrieved==0) && maxLocalLocationsImmunized==0) && signature->getWeight()>=0)
 	{
 		if(retrievalId > 0)
 		{
@@ -2445,7 +2445,7 @@ bool Rtabmap::process(
 			}
 		}
 
-		if(!(_memory->allNodesInWM() && maxLocalLocationsImmunized == 0))
+		if(!((_memory->allNodesInWM() || _maxLocalRetrieved==0) && maxLocalLocationsImmunized==0))
 		{
 			// immunize the path from the nearest local location to the current location
 			if(immunizedLocally < maxLocalLocationsImmunized &&
--- a/corelib/src/Rtabmap.cpp
+++ b/corelib/src/Rtabmap.cpp
@@ -2454,7 +2454,7 @@ bool Rtabmap::process(
 			}
 		}
 
-		if(!((_memory->allNodesInWM() || _maxLocalRetrieved==0) && maxLocalLocationsImmunized==0))
+		if(!(_memory->allNodesInWM() && maxLocalLocationsImmunized==0))
 		{
 			// immunize the path from the nearest local location to the current location
 			if(immunizedLocally < maxLocalLocationsImmunized &&
--- a/corelib/src/Rtabmap.cpp
+++ b/corelib/src/Rtabmap.cpp
@@ -2318,7 +2318,13 @@ bool Rtabmap::process(
 							//immunized locations in the neighborhood from being transferred
 							if(immunizedLocations.insert(iter->first).second)
 							{
-								++immunizedGlobally;
+								// Count only non-intermediate nodes (intermediate nodes are still
+								// immunized but don't consume the immunization budget/statistic).
+								const Signature * sImmunized = _memory->getSignature(iter->first);
+								if(sImmunized == 0 || sImmunized->getWeight() >= 0)
+								{
+									++immunizedGlobally;
+								}
 							}
 
 							//UDEBUG("nt=%d m=%d immunized=1", iter->first, iter->second);
@@ -2421,9 +2427,12 @@ bool Rtabmap::process(
 					distanceSoFar += _path[i-1].second.getDistance(_path[i].second);
 				}
 
-				if(_memory->getSignature(_path[i].first) != 0)
+				const Signature * sPath = _memory->getSignature(_path[i].first);
+				if(sPath != 0)
 				{
-					if(immunizedLocations.insert(_path[i].first).second)
+					// Count only non-intermediate nodes (intermediate nodes are still
+					// immunized but don't consume the immunization budget/statistic).
+					if(immunizedLocations.insert(_path[i].first).second && sPath->getWeight() >= 0)
 					{
 						++immunizedLocally;
 					}
@@ -2513,7 +2522,13 @@ bool Rtabmap::process(
 								{
 									if(immunizedLocations.insert(iter->first).second)
 									{
-										++immunizedLocally;
+										// Count only non-intermediate nodes (intermediate nodes are still
+										// immunized but don't consume the immunization budget/statistic).
+										const Signature * sLocal = _memory->getSignature(iter->first);
+										if(sLocal == 0 || sLocal->getWeight() >= 0)
+										{
+											++immunizedLocally;
+										}
 									}
 									//UDEBUG("local node %d on path immunized=1", iter->first);
 								}
@@ -2573,7 +2588,9 @@ bool Rtabmap::process(
 					}
 					if(!_memory->isInSTM(s->id()) && immunizedLocally < maxLocalLocationsImmunized)
 					{
-						if(immunizedLocations.insert(s->id()).second)
+						// Count only non-intermediate nodes (intermediate nodes are still
+						// immunized but don't consume the immunization budget/statistic).
+						if(immunizedLocations.insert(s->id()).second && s->getWeight() >= 0)
 						{
 							++immunizedLocally;
 						}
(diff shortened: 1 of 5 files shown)


matlabbe:
What is the use case that is failing for you?
```
Rtabmap/CreateIntermediateNodes=true
Rtabmap/MemoryThr=?
RGBD/LocalImmunizationRatio=0
RGBD/MaxLocalRetrieved=0
Rtabmap/MaxRetrieved=0
... ???
```

I think in https://github.com/introlab/rtabmap/pull/1687 I made sure it was doing Transfer/Retrieval (what is expected from the papers) like if intermediate nodes were not used (the count of retrieved nodes or transferred nodes are based on the non-intermediate nodes). I can doublecheck though. For the recording version with `Rtabmap/MemoryThr=1 or 2`, why even using intermediate nodes? `Rtabmap/DetectionRate` can be set to 0 to record everything (with `Kp/MaxFeatures=-1`). I can retest intermediate nodes in data recording mode, though it was not supposed to be used together.

Not convinced we need to change the way working memory size is used (and counting the virtual or not). In particular, for thousands of nodes, we may spend significant time in `getWorkingMemSize()`, if it is also called multiple times per iteration. I would rater track with a local variable the number of intermediate nodes in Memory during update/retrieval/transfer. However, the only place I see we would want to know about the intermediate nodes is to avoid to scale manually `Rtabmap/MemoryThr` parameter. 

For `RGBD/LocalImmunizationRatio`, I didn't check explicitly, but it should behave similarly having intermediate nodes or not, because it is a ratio and the numbers compared would include the intermediate nodes if used, or not if not used.

borongyuan:
> What is the use case that is failing for you?

I tested it using the latest release version of RTAB-Map (v0.23.7), with the following configuration.
```
Kp\BadSignRatio = 0.0
Mem\BinDataKept = true
Mem\RecentWmRatio = 0.0
Mem\RehearsalSimilarity = 1.0
Mem\STMSize = 1
RGBD\AngularUpdate = 0
RGBD\LinearUpdate = 0
RGBD\LocalImmunizationRatio = 0.0
RGBD\LoopCovLimited = true
RGBD\MaxLocalRetrieved = 0
RGBD\ProximityBySpace = false
Rtabmap\CreateIntermediateNodes = true
Rtabmap\DetectionRate = 1.0
Rtabmap\MaxRetrieved = 0
Rtabmap\MemoryThr = 2
Vis\PnPSplitLinearCovComponents = true
Vis\PnPVarianceMedianRatio = 2
g2o\PixelVariance = 2.0
```
The statistics show that Immunized locally is 0, but Immunized globally increases, and WM size also increases. Fortunately, they do not appear to grow indefinitely under this configuration, but rather fluctuate within the range of tens to hundreds. Since we set `Mem\RehearsalSimilarity = 1.0` and `Mem\STMSize = 1`, the input data frame rate is 20 Hz, and the STM size in the statistics is 20 or 21, which is as expected. However, if we change Mem\STMSize back to 10 and enable rehearsal, the size of the STM will continue to grow indefinitely.
截图 2026-07-10 12-12-39

> I think in #1687 I made sure it was doing Transfer/Retrieval (what is expected from the papers) like if intermediate nodes were not used (the count of retrieved nodes or transferred nodes are based on the non-intermediate nodes).

There should be no problem with counting in Transfer/Retrieval processing. Transfer/Retrieval skips processing intermediate nodes directly and instead processes the intermediate nodes connected to them while processing regular nodes, which is a very good approach. The problem lies in the calculation of `maxLocalLocationsImmunized`. Because `_memory->getWorkingMem().size()` includes intermediate nodes, maxLocalLocationsImmunized is magnified by about 20 times during our runtime. Then, comparing the actual number of regular nodes processed with this will result in an unexpected number of regular nodes being processed.
https://github.com/introlab/rtabmap/blob/bc724a21254a055cee7581d764e4513a0d6fa8ce/corelib/src/Rtabmap.cpp#L2249-L2263

Since `_memory->allNodesInWM()` is false, the following code will always be executed. When calling `_memory->getNeighborsId()`, even though the passed-in `_maxRetrieved` is 0, this only prevents the node from being retrieved from the LTM. Since the neighbors are already in the WM, they are immunized.
https://github.com/introlab/rtabmap/blob/bc724a21254a055cee7581d764e4513a0d6fa8ce/corelib/src/Rtabmap.cpp#L2267-L2294

> For the recording version with `Rtabmap/MemoryThr=1 or 2`, why even using intermediate nodes? `Rtabmap/DetectionRate` can be set to 0 to record everything (with `Kp/MaxFeatures=-1`).

This is still due to the reason I mentioned in #1697. Some users are now using RTAB-Map for embodied AI data acquisition. The trajectories they are collecting are getting longer and longer, from tens of seconds at the beginning to tens of minutes now. They even collect tracks from multiple devices simultaneously and then merge them using rtabmap-reprocess. If we treat all nodes as regular nodes, the amount of data will be enormous. This also makes offline loop closure detection and merging much slower. Furthermore, the behavior of Bayesian filtering and loop closure detection can even change due to the high similarity between neighbor nodes. Enabling intermediate nodes simply makes the recorded trajectories more continuous; they participate in graph optimization but not in loop closure detection. I noticed that [commit d069bec](https://github.com/introlab/rtabmap/commit/d069becc724905d6465a8758c40dae88e13bb9e8) added the setIMU() method, which may even be used to record higher frequency trajectories in the future. In our custom VIO, although the image data is only 20 Hz, we will provide high-frequency pose output according to the IMU frequency, which is usually 200 Hz. We will obtain neural network features from the OAK camera. Both the front-end and back-end will skip feature detection, so setting `Kp/MaxFeatures` have no effect. Feature data is transmitted to RTAB-Map along with image data via rtabmap_msgs/RGBDImage. This eliminates the need to enable neural network support when compiling RTAB-Map, and prebuilt RTAB-Map distributed by ROS can be directly used.

> Not convinced we need to change the way working memory size is used (and counting the virtual or not). In particular, for thousands of nodes, we may spend significant time in `getWorkingMemSize()`, if it is also called multiple times per iteration. I would rater track with a local variable the number of intermediate nodes in Memory during update/retrieval/transfer.

In our current applications, either for general applications where intermediate nodes are disabled and `Rtabmap\MemoryThr` is set to 1000, or for data recording where intermediate nodes are enabled and `Rtabmap\MemoryThr` is set to 1 or 2. Calling `getWorkingMemSize(true)` will perform an iteration, but the performance impact should not be significant. In the new C++ standard, `std::count_if` can also be used, but its time complexity is also O(N), which is not much different from direct traversal.

borongyuan:
> The statistics show that Immunized locally is 0, but Immunized globally increases, and WM size also increases. Fortunately, they do not appear to grow indefinitely under this configuration, but rather fluctuate within the range of tens to hundreds.

Some additional analysis on this:
In RETRIEVAL 1/3, the neighbors of `retrievalId` are inserted into `reactivatedIds`.
In RETRIEVAL 2/3, since `_maxLocalRetrieved` is 0, `retrievalLocalIds.size() < _maxLocalRetrieved`, no new data is inserted into `reactivatedIds`.
In RETRIEVAL 3/3, `reactivatedIds.size() > 0`, and `_memory->reactivateSignatures()` is called. Since a `maxLoaded` value of 0 effectively means unlimited reactivation/reloading, all `reactivatedIds` are added to `immunizedLocations`.
https://github.com/introlab/rtabmap/blob/bc724a21254a055cee7581d764e4513a0d6fa8ce/corelib/src/Memory.cpp#L6966-L6985

corelib/src/Rtabmap.cpp
```diff
@@ -2447,7 +2445,7 @@ bool Rtabmap::process(
 			}
 		}
 
-		if(!(_memory->allNodesInWM() && maxLocalLocationsImmunized == 0))
+		if(!((_memory->allNodesInWM() || _maxLocalRetrieved==0) && maxLocalLocationsImmunized==0))
```
matlabbe: Is this change necessary for some cases? Because we may want to use `nearNodesByDist` (inside that if) to update the age of the nodes when `_maxLocalRetrieved` is 0.

matlabbe:
I played with these changes for a while, looks good! I just improved the `getWorkingMemSize()` complexity when requesting without intermediate nodes. There was a loop `O(N)` on the working memory size map, and then a lookup in a map, producing a total of `O(N log (N))`. With a cached variable to track the number of intermediate nodes, we can have O(1).

You may doublecheck if everything is still working for you, then I'll merge.

corelib/src/Rtabmap.cpp
borongyuan: You are right, this change is not necessary. If _maxLocalRetrieved is 0, the subsequent code will also avoid local retrieval and immunization.

borongyuan:
I have revised the condition for counting intermediate nodes. This code will treat invalid nodes with a weight of -9 as intermediate nodes. This can even cause _stMemIntermediateNodesCount to become negative. The reason is that setWeight(-9) appears not only in Memory::moveToTrash(), but also in Memory::rehearsalMerge(). If rehearsal is enabled, there will already be invalid nodes in the STM before calling Memory::moveToTrash().
https://github.com/introlab/rtabmap/blob/531b8a43031a85d5d97d1192093d203e7a3ad54b/corelib/src/Memory.cpp#L2759-L2776
截图 2026-07-13 22-06-15

matlabbe:
LGTM!

PR #1492, opened 2025-04-25
Adding doc and tests

matlabbe:
What this PR brings:
* Tests!
* API documentation
* Coverage report artifacts
* Doxygen artifacts
* ~Sphinx (ros2doc) artifacts~ will do rosdoc2 only for `rtabmap_ros` packages
* `gh_pages` now generated by github actions, the main page will have to be modified under `website` folder on master branch

Breaking changes:
* ~USemaphore's acquire/acquireTry/release and UMutex's lock/lockTry/unlock are all now returning true or false (instead of nebulous error code), so it means if someone was calling `if(mutex_.lockTry() == 0)`, they should now do `if(mutex_.lockTry())` (inverse), meaning we could lock the mutex.~ Actually, now only `USempahore::acquireTry()` on linux is inverted (standardized with the other ones).
* For parameter names that have change, the downstream code would have compilation errors.
* **PythonInterface**: If rtabmap is built with python support, you still have to instantiate the `PythonInterface` on the main thread (see PythonInterface documentation). However, you have to change:
   ```diff
   #include <rtabmap/core/Version.h>
   #ifdef RTABMAP_PYTHON
   #include <rtabmap/core/PythonInterface.h>
   #endif
   
   int main(int argc, char * argv[]) {
   
   #ifdef RTABMAP_PYTHON
   - rtabmap::PythonInterface pythonInterface;
   + rtabmap::PythonInterface::instance("my_app_main");
   #endif
   
   [...]
   return 0;
   }
   ```

PR #1741, opened 2026-07-23
OptimizerG2O: track BA outliers per observation

### LINKED ISSUE #1652: OptimizerG2O::optimizeBA(): should we track edge outliers instead of 3D point outliers?
I think here we should track 2D edge outliers instead of 3D points. Only when the 3D points with all 2D references as outliers would be reset to its original position. optimizeBA() would return not a list of 3D points outliers, but a corresponding list of 2D outliers (matching the `wordReferences` map instead)

https://github.com/introlab/rtabmap/blob/b47c4093a6014ae2ae897b0fcc6af0cc45e036b1/corelib/src/optimizer/OptimizerG2O.cpp#L1962-L2000

Ref: http://official-rtab-map-forum.206.s1.nabble.com/RTAB-Map-and-COLMAP-tp9344p11536.html

admin123-abc:
Fixes #1652.

A rejected reprojection edge previously marked the entire landmark as an outlier. As a result, one bad observation could restore a landmark to its input position even when its other projections were valid.

This PR tracks BA outliers per `(wordId, poseId)` observation, matching `wordReferences`. Callers now remove only the rejected observation for their target pose. A landmark falls back to its input estimate only after all of its projection edges have been rejected.

A deterministic g2o regression covers:

- a landmark with eight valid projections and one bad projection, which must remain optimized and report only the bad observation;
- a control landmark with two rejected projections, which must fall back to its input position.

Validation:

- Built `rtabmap_core` with g2o + Ceres and ran `optimizer_g2o_ba_outliers` successfully 20 times.
- Built `rtabmap_core` with g2o disabled and Ceres enabled.
- Added the test to the Ubuntu 24.04 CMake CI job.


matlabbe:
on-hold till https://github.com/introlab/rtabmap/pull/1492 ([gtest](https://github.com/introlab/rtabmap/tree/gtest) branch) (having testing framework) is merged.

Otherwise, LGTM! You can target `gtest` branch instead of `master` to fix any merge conflicts now, otherwise I'll do it after `gtest` is merged to `master`.

admin123-abc:
Thanks! I'll wait for #1492 to be merged and handle any conflicts afterward if needed.

matlabbe:
I updated it with master and refactored gtsam and ceres to do the same. Once CI passes, I'll merge it.

PR #1753, opened 2026-08-24
Fix an issue with local BA and improve on-demand signature loading

borongyuan:
Hi,

I discovered an issue introduced in #1569. If a signature only has 2D features and no 3D features, it will be incorrectly allowed to be used to build a local feature map, which will cause segfault when `s->getWords3()[jter->second]` is called.

After fixing this issue, I discovered that #1569 also did not perform on-demand loading of the neighbors of fromS. As you mentioned in your comments, `Mem/LoadVisualLocalFeaturesOnInit` may cause `RGBD/LocalBundleOnLoopClosure` not to work properly. So I added this part of the implementation, which is similar to the on-demand loading of formS. Generally, only 3-4 nodes participate in local BA: fromS, toS, and the preceding and following neighbors of fromS. Therefore, loading 1-2 more signatures should not incur significant overhead.

Additionally, if fromS and toS are linked by `Link::kNeighborMerged`, we should also allow local BA between them if fromS has other neighbors and other conditions are met. If a region is scanned repeatedly, after graph reduction, a node may be linked to all its neighboring nodes as `Link::kNeighborMerged`. It would be not ideal if we only allowed registration refinement for `Link::kNeighborMerged`, since there were indeed one or more loop closures here. Even for loop closure links that have already been merged, performing a local BA is still helpful.


matlabbe:
Looks good to me! Letting CI catch up on the merge from master and will merge it afterwards...

PR #1757, opened 2026-08-26
Parallelize cloud generation in export/view clouds dialog and rtabmap-export

torjusti:
This PR multi threads the cloud generation when exporting from the DB viewer or using the CLI export tool. Only the cloud generation is affected, the cloud assembly is still sequential.

With 32 threads it's a 16x speedup for me, very significant as I have databases with lots of nodes in them and the cloud export takes a long time. The output should be identical to the existing implementation.

The diff is a bit noisy because all the properties needed for cloud exports have been extracted to a separate data object. That's done because I'm worried about interacting with the Qt widget values on a different thread, but maybe it's fine. Open to suggestions for better ways of doing this!


tools/Export/main.cpp
```diff
@@ -1832,56 +1895,80 @@ int main(int argc, char * argv[])
...
-			if(percent != lastPercent)
+			loadNode(nodes[chunkStart+i].first, nodes[chunkStart+i].second, chunkData[i]);
+		}
+
+		for(size_t i=0; i<chunkNodes; ++i)
+		{
+			flushNode(nodes[chunkStart+i].first, nodes[chunkStart+i].second, chunkData[i]);
+			chunkData[i] = NodeExportData();
+
+			if(optimizedPoses.size() >= 500)
```
matlabbe, line 1990: To help understanding, if I have 16 cores, does it do:
1. Load/create cloud of nodes 1->16 in parallel
2. Append clouds 1->16 sequentially to the assembled cloud
3. Repeat 1. 2. with nodes 17->32
4. Repeat 1. 2. with nodes 33->48
5. and so on?

So we have at maximum 16 clouds in RAM to assemble, then we clear them.

guilib/src/ExportCloudsDialog.cpp
```diff
@@ -3990,144 +3925,426 @@ std::map<int, std::pair<pcl::PointCloud<pcl::PointXYZRGBNormal>::Ptr, pcl::Indic
...
+		const std::map<int, LaserScan> & cachedScans,
+		const ParametersMap & parameters,
+		bool & has2dScans,
+		bool & scansHaveRGB) const
+{
+	scansHaveRGB = false;
+	has2dScans = false;
+	std::map<int, std::pair<pcl::PointCloud<pcl::PointXYZRGBNormal>::Ptr, pcl::IndicesPtr> > clouds;
+
+	CloudGenParams p;
```
matlabbe: I wonder if `CloudGenParams` is necessary, reading checkbox value or spinbox value from multi threads should be safe. The only thing Qt doesn't like is when we do UI related functions in other threads than Main Thread.

tools/Export/main.cpp
torjusti: Yes, essentially. The "chunk size" (number of nodes processed per iteration before we assemble) is set to 4x the core count currently, my thinking being that it could help saturate the cores better. But yes that's the overall idea.

At least for the rtabmap-export tool. The DB viewer is different, that app keeps a list of all exported clouds in memory for further processing (existing behavior). So in that implementation the threads just keep going without freeing anything.

guilib/src/ExportCloudsDialog.cpp
torjusti: Yeah, turned out to work fine without it. I've pushed changes to not do that now

fallenmi (approved):
Disclosure: I used Codex to assist with this review and verified the result against exact commit `be7714e7b8b5a0d14e52bd2f2ad5cfb7db2ea880`.

I traced both parallel refactors against base `d4940634`: the GUI keeps subtraction filtering on its sequential path and publishes every other result atomically before flushing it in node order, while the CLI performs only per-node loading/generation in parallel and preserves node order for cloud assembly, pose/GPS/grid accumulation, counters, and progress.

A focused OpenMP oracle exercised 3,000 randomized 1/2/4/8-thread ordering and cancellation cases in both normal and ASan/UBSan builds; all passed. The current GitHub merge `b51f4dc2` contains the same versions of all three changed files, and the exact-head upstream matrix is green across 25 executed jobs, with one dependency job intentionally skipped. I found no blocking issue.

matlabbe:
I reviewed more deeply and did some minor changes (matching commits above):
* [Simplified NodeExportData by holding Signature directly. Added --threads option (default max cores) for CLI and UI.](https://github.com/introlab/rtabmap/pull/1757/changes/4caff5c93b12714dbf734ba7de2f6aece0e6c7a9)
* [Parallelized texturing phase the same way than cloud generation](https://github.com/introlab/rtabmap/pull/1757/changes/672f0131f3cccf5d5aca22f63cbdad53e869ca56) (New thing, I could use easily the same pattern you used for this one, saving significantly more time)
* [Fixed not cancelable (right away) cloud generation in UI](https://github.com/introlab/rtabmap/pull/1757/changes/22f778d1924b68c915893600cb096a478c097fd2)

Experiment: I used combined `merged.db` from https://github.com/introlab/rtabmap/wiki/Multi-Session-Mapping-with-RTAB-Map-Tango, then with `rtabmap-export` (on a 16 cores computer):
| Step                 | Before | After  | Speedup |
|----------------------|-------:|-------:|--------:|
| Cloud generation     |    7 s |  0.7 s |     10x |
| Mesh texturing       |   35 s |    9 s |    3.9x |
| **Total**            | **42 s** | **9.7 s** | **4.3x** |

Commands used:
```bash
# Assemble
$ rtabmap-reprocess --Optimizer/GravitySigma 0.3 --Mem/UseOdomGravity true "chalet1.db;chalet2.db;chalet3.db;chalet4.db" merged.db

# Processing with 1 thread (like before)
rtabmap-export --threads 1 --texture --texture_size 4096 --texture_range 3 --poisson_depth 9 --max_polygons 500000 --multiband --color_radius 0 --no_clean   merged.db

# Processing with 16 threads (default is 0 -> auto)
rtabmap-export --threads 0 --texture --texture_size 4096 --texture_range 3 --poisson_depth 9 --max_polygons 500000 --multiband --color_radius 0 --no_clean  merged.db
```

One downside for texturing is that there is a 50 MB overhead per thread. So 16 threads means 15*50MB=750MB overhead in RAM. For only cloud generation, the penalty in RAM usage is way less.

tools/Export/main.cpp
matlabbe: After playing with it, indeed having x4 helps to be slightly faster.

matlabbe (approved):

PR #1760, opened 2026-09-07
 VWDictionary: use multi-core FLANN kNN search

torjusti:
This PR enabled multi threaded knnSearch when searching in the dictionary. On my machine it gives a modest speedup to rtabmap-reprocess.

Some data when reprocessing 8000 nodes of a database, 500 SuperPoint features per node (times are only for the search step):

| Dictionary size | cores=1 | cores=all | Speedup |
|---|---:|---:|---:|
| <50k | 7.4 ms | 0.78 ms | ~9× |
| 50–200k | 12.7 ms | 1.13 ms | ~11× |
| 200–500k | 16.7 ms | 1.34 ms | ~12× |
| 500k–1M | 20.0 ms | 1.55 ms | ~13× |
| ≥1M | 23.6 ms | 1.87 ms | ~13× |

Overall: **20.1** to **1.60 ms/node** on average (~12.5×)

Synthetic benchmark across different dictionaries and feature counts:

| Dictionary | Dim | cores=1 | cores=4 | cores=all | Speedup |
|---|---:|---:|---:|---:|---:|
| 100k | 128 | 5.5 ms | 1.9 ms | 0.45 ms | ~12× |
| 100k | 256 | 6.8 ms | 2.0 ms | 0.61 ms | ~11× |
| 500k | 128 | 8.0 ms | 2.3 ms | 0.72 ms | ~11× |
| 500k | 256 | 10.2 ms | 2.9 ms | 0.95 ms | ~11× |

No parameter currently, it defaults to max threads. Happy to change it to be a parameter if wanted.


corelib/src/VWDictionary.cpp
```diff
@@ -1074,7 +1074,7 @@ std::list<int> VWDictionary::addNewWords(
 
 		if(isFlannStrategy(_strategy))
 		{
-			_flannIndex->knnSearch(descriptors, results, dists, k, KNN_CHECKS);
+			_flannIndex->knnSearch(descriptors, results, dists, k, KNN_CHECKS, 0.0f, true, 0);
```
matlabbe: I would make an option to opt-in. The global parameter could be called `Kp/FlannThreads`, default to 1, 0=auto.

corelib/src/FlannIndex.cpp
```diff
@@ -930,6 +934,15 @@ void FlannIndex::knnSearch(
...
 
 	rtflann::SearchParams params = rtflann::SearchParams(checks, eps, sorted);
+	if(cores <= 0)
+	{
+#ifdef _OPENMP
+		cores = omp_get_max_threads();
+#else
+		cores = 1;
+#endif
+	}
```
matlabbe: ~~I don't think this is necessary, based on the doc of SearchParams, 0=auto... you may double check the internal implementation.~~

~~You may only check for <0 and set to zero if so. Not sure if negative is handled in flann.~~

EDIT: Actually, based on the official [OpenMP doc](https://www.openmp.org/spec-html/5.0/openmpse14.html#x54-800002.6), cores should be positive:
> The num_threads expression must evaluate to a positive integer value. 

However, not 100% sure if 0 is considered positive. On linux, 0 seems to automatically give max cores, though I could not test on Windows. Lets keep that way for now.

corelib/src/VWDictionary.cpp
matlabbe: That could be the reason why the windows test is timing out 1800 sec, I checked the last success on that job was 1670.17 sec. I triggered another build to see if it is consistently timing out with all cores.

```
21: [          ] Replay summary: framesRead=935 framesProcessed=935 framesIntermediate=0 loops=224
 loopsRejected=449 proximity=0 localGraph=249 globalGraph=935 rmse=-1m posterior(avg/min/
max)=0.333981/0/9.8747ms wall=109.326s
[...]
18/22 Test #21: test_rtabmap_integration .........***Timeout 1800.43 sec
```

Last green windows build:
```
 21: [          ] Replay summary: framesRead=935 framesProcessed=935 framesIntermediate=0 loops=217
 loopsRejected=446 proximity=0 localGraph=254 globalGraph=935 rmse=-1m posterior(avg/min/
max)=0.38676/0/11.7295ms wall=70.7677s
[...]
18/22 Test #21: test_rtabmap_integration .........   Passed  1670.17 sec
```

corelib/src/FlannIndex.cpp
torjusti: Negative values crashes, 0 it does seem to give max cores but I also haven't tested that on Windows

corelib/src/VWDictionary.cpp
torjusti: Added param defaulting to 1 core

matlabbe: Perfect, thanks!

matlabbe:
Not sure what is the deal with the windows tests failing, I'll take care of that. Otherwise, the PR looks good to me.

matlabbe (approved):

PR #1766, opened 2026-09-12
Memory: reuse compressed image/depth blobs when pixels are unchanged

torjusti:
Still chasing optimizations but they are getting harder to find :slightly_smiling_face: 

rtabmap-reprocess currently re-compresses every image, this skips the compression step if the image has not been changed. It only saves a couple milliseconds for me per frame so might not be worth the additional complexity.

There is also the additional benefit that we don't damage images with repeated JPEG compression if reprocessing reprocessed databases, although I don't think it amounts to much


matlabbe:
That makes sense, thanks for the PR. Gonna fix the flaky test and merge this.

